PR #2, opened 2019-04-29
Create common cmake modules for dependency resolution [ESD-1245]

woodfell:
This PR implements some common cmake functions used for the dependency resolution and provides standard Find* modules for common packages.

The two main new modules are:

GenericFindDependency
A function to search for a dependency. Searches in either bundled source code or system libraries. Configurable through either function parameters or global variables defined on the command line or elsewhere at configure time. Don't call directly, call from a Find* style cmake module. 

SwiftCmakeOptions
Sets up and configures several generic options to enable or disable package features such as unit tests, documentation, or example code.

GenericFindDependency.cmake and SwiftCmakeOptions.cmake contain header with documentation of parameters and usage. The PR also includes several Find* modules for commonly used packages including all swift repositories. See other repos for examples of SwiftCmakeOptions.

As part of this work the new modules are being imported to other swift repositories.
Related PRs:

libsbp - https://github.com/swift-nav/libsbp/pull/692
librtcm - https://github.com/swift-nav/librtcm/pull/85
libswiftnav - https://github.com/swift-nav/libswiftnav/pull/85
gnss-converters https://github.com/swift-nav/gnss-converters/pull/193
libsettings - https://github.com/swift-nav/libsettings/pull/37
starling - https://github.com/swift-nav/starling/pull/2273
albatross - https://github.com/swift-nav/albatross/pull/126
orion_proto - https://github.com/swift-nav/orion_proto/pull/31
orion-engine - https://github.com/swift-nav/orion-engine/pull/120
orion - https://github.com/swift-nav/orion/pull/400


FindFastCSV.cmake
```diff
@@ -0,0 +1,15 @@
+include("GenericFindDependency")
+GenericFindDependency(
+    TargetName "fast-csv"
+    SourceDir "fast-cpp-csv-parser"
+    )
+
+if(NOT CMAKE_CROSSCOMPILING OR THIRD_PARTY_INCLUDES_AS_SYSTEM)
```
benjaminaltieri: this seems like a pattern we could generalize? possibly add `IncludeAsSystem [FORCE] TARGET` in the `GenericFindDependency` call which could handle the default behavior of switching includes to `SYSTEM` if `NOT CMAKE_CROSSCOMPILING OR THIRD_PARTY_INCLUDES_AS_SYSTEM` or explicitly force system includes if needed

FindGnss-converters.cmake
```diff
@@ -0,0 +1,5 @@
+include("GenericFindDependency")
+GenericFindDependency(
+    TargetName gnss_converters
+    SourceDir "gnss-converters/c"
```
benjaminaltieri: should this be `SourceDir "c"`?

benjaminaltieri:
Looks great so far @woodfell! Documentation is 💯

GenericFindDependency.cmake
```diff
@@ -0,0 +1,226 @@
...
+    endif()
+  endif()
+
+  foreach(LOCATION ${x_Locations})
+    if (${LOCATION} STREQUAL "source")
+      # set defaults
+      set(x_SourceSearchPaths "")
+      if(NOT x_SourceDir)
+        list(APPEND x_SourceSearchPaths "${CMAKE_CURRENT_SOURCE_DIR}/third_party/${CMAKE_FIND_PACKAGE_NAME}")
+        list(APPEND x_SourceSearchPaths "${CMAKE_CURRENT_SOURCE_DIR}/third_party/${x_TargetName}")
```
jbangelo: Any particular reason the source search path prioritizes `CMAKE_CURRENT_SOURCE_DIR` over  `PROJECT_SOURCE_DIR`? It doesn't really matter to me, just curious. I guess this prefers "closer" directories first which makes sense.

FindCheck.cmake
```diff
@@ -0,0 +1,55 @@
...
+#  Copyright (c) 2007-2009 Bjoern Ricks  <bjoern.ricks@gmail.com>
+#
+#  Redistribution and use is allowed according to the terms of the New
+#  BSD license.
+#  For details see the accompanying COPYING-CMAKE-SCRIPTS file.
+
+
+INCLUDE( FindPkgConfig )
+
+IF ( Check_FIND_REQUIRED )
```
martin-swift: can we go lowercase to be consistent with the other files?

FindCheck.cmake
```diff
@@ -0,0 +1,55 @@
...
+	SET( _pkgconfig_REQUIRED "REQUIRED" )
+ELSE( Check_FIND_REQUIRED )
```
martin-swift: I have been trying to move the cmake away from including the variable where it is not actually being tested.  (In my opinion, this is a bug in the CMake syntax to allow this...)

martin-swift: had a quick go in my editor to see how it would look:
```
include( FindPkgConfig )

if ( Check_FIND_REQUIRED )
    set( _pkgconfig_REQUIRED "REQUIRED" )
else()
    set( _pkgconfig_REQUIRED "" )
endif()

if ( CHECK_MIN_VERSION )
    pkg_search_module( CHECK ${_pkgconfig_REQUIRED} check>=${CHECK_MIN_VERSION} )
else ( CHECK_MIN_VERSION )
    pkg_search_module( CHECK ${_pkgconfig_REQUIRED} check )
endif ( CHECK_MIN_VERSION )

# Look for CHECK include dir and libraries                                                                             
if( NOT CHECK_FOUND AND NOT PKG_CONFIG_FOUND )

    find_path( CHECK_INCLUDE_DIRS check.h )
    find_library( CHECK_LIBRARIES NAMES check )

    if ( CHECK_INCLUDE_DIRS AND CHECK_LIBRARIES )
        set( CHECK_FOUND 1 )
        if ( NOT Check_FIND_QUIETLY )
            message ( STATUS "Found CHECK: ${CHECK_LIBRARIES}" )
        endif ( )
    else ()
        if ( Check_FIND_REQUIRED )
            message( FATAL_ERROR "Could NOT find CHECK" )
        else ()
            if ( NOT Check_FIND_QUIETLY )
                message( STATUS "Could NOT find CHECK" )
            endif ( )
        endif ( )
    endif ()
endif()

# Hide advanced variables from CMake GUIs                                                                              
mark_as_advanced( CHECK_INCLUDE_DIRS CHECK_LIBRARIES )
```

FindCheck.cmake
```diff
@@ -0,0 +1,55 @@
...
+	SET( _pkgconfig_REQUIRED "" )
+ENDIF ( Check_FIND_REQUIRED )
+
+IF ( CHECK_MIN_VERSION )
```
martin-swift: So why CHECK all caps in some variables and `Check` in others?  Is there a convention I have overlooked?

FindCheck.cmake
martin-swift: Could also kill a level of nested of those nested `if()` statements if you made a wrapper for `message()` that checks `Check_FIND_QUIETLY`

FindCheck.cmake
benjaminaltieri: FWIW, per reviewing `FindCheck.cmake`, I believe it was taken whole from `libsbp` I'm assuming for expediency's sake: https://github.com/swift-nav/libsbp/blob/master/c/cmake/FindCheck.cmake

FindCheck.cmake
```diff
@@ -0,0 +1,55 @@
...
+		SET( CHECK_FOUND 1 )
+		IF ( NOT Check_FIND_QUIETLY )
+			MESSAGE ( STATUS "Found CHECK: ${CHECK_LIBRARIES}" )
+		ENDIF ( NOT Check_FIND_QUIETLY )
+	ELSE ( CHECK_INCLUDE_DIRS AND CHECK_LIBRARIES )
+		IF ( Check_FIND_REQUIRED )
+			MESSAGE( FATAL_ERROR "Could NOT find CHECK" )
+		ELSE ( Check_FIND_REQUIRED )
+			IF ( NOT Check_FIND_QUIETLY )
+				MESSAGE( STATUS "Could NOT find CHECK" )
```
martin-swift: emacs' CMake mode uses 2 spaces for indenting by default (and Swift also uses 2 spaces for indentation in C/C++), so that's what we have been using thus far.  It's not particularly hard for me to change the default indentation, but if we are going to change it, let's do it intentionally.  Thoughts?

martin-swift: ah, that's why everything is so different - looks like this one is not of Swift provenance.  (should have read the header first!)

benjaminaltieri:
@martin-swift RE my response to your review comments above, I think there's some management of expectations to be done for this repo in terms of what uses common methods vs taking cmake modules provided for third party repos as they are. I don't know what the right mix is per se, but figured that I'd call it out as an open question as to how much modification is really necessary for such open-sourced reuse.

FindEigen.cmake
```diff
@@ -0,0 +1,19 @@
+if(TARGET eigen)
```
benjaminaltieri: is there a reason not to use the `GenericFindDependency` pattern here?

FindCheck.cmake
benjaminaltieri: heh yeah I think it's the only one that's imported like that, which makes it stick out like a sore thumb!

FindEigen.cmake
jbangelo: I think this was pulled in from the starling repo. It's a header only library, so it was simple enough to redefine its target in CMake.

It would be an interesting experiment to use `GenericFindDependency` on Eigen. Its CMake files do a lot of checks of the target system and might enable some optimizations that we're missing out on.

woodfell: We use eigen as a header only library. Initially I did use the GenericFindDependency pattern which ended up actually compiling the library, but that resulted in no end of compiler errors due to -Werror and other options. In this case it was better to just keep the separate interface target.

FindCheck.cmake
woodfell: Absolutely right. FindCheck.cmake was actually in several repos, not just libsbp. I copied it in here simply to avoid duplication. I'm happy to revisit and tidy up if that's the feeling, but since this ticket is already gigantic let's do it elsewhere.

GenericFindDependency.cmake
woodfell: Yes, there is one example, although I can't remember where, where third_party appeared underneath a test directory. CMAKE_CURRENT_SOURCE_DIR seems to cover the bases. CMAKE_CURRENT_LIST_DIR might be a suitable alternative, we could experiment with that.

FindCheck.cmake
martin-swift: Agreed.  Can probably even pull it out of the other repos once this goes in, no?

FindGnss-converters.cmake
woodfell: No, we have a problem in several repos of mixing up hyphen and underscore. orion_proto/orion-proto is another example of this. Without the SourceDir option the function would search in 'third_party/gnss_converters' which doesn't exist.

FindFastCSV.cmake
woodfell: Yep, good point

FindCheck.cmake
woodfell: Already done that, the PRs which are open against all the other repos remove FindCheck.cmake and import it from this common cmake module repo instead.

benjaminaltieri: Just to link everything together can you add a reference to this PR (either in the PR description or via comment) in the other PRs that are open which incorporate these modules? that would make it easy to see the various examples of usage

martin-swift: are those pulls ready for review, or still WIP?

benjaminaltieri: ^^ @woodfell

woodfell: Everything is ready to go, with the exception of starling which has some failing unit tests I'm still looking in to. We had a brief discussion yesterday in our standup about how to manage all these merges, I think we will start off with a few low level projects like libsbp and libswiftnav, maybe gnss-converters, verify them first, and then start working our way up to ultimately conclude with orion.

FindGnss-converters.cmake
martin-swift: And not just the hyphen/underscore dissonance, no?  in this case gnss-converters is a multi-language repo so it has a slightly different structure from the C-only repos.

woodfell: Right. libsbp is another example of that.

martin-swift: ok, I see your vision now; you do a bunch of permutations to catch all the weird cases...

woodfell: Pretty much, search all the most likely locations by default with the ability to massage if required

FindCheck.cmake
martin-swift: great.  can you tag me on those pulls?

GenericFindDependency.cmake
```diff
@@ -0,0 +1,226 @@
+#
```
martin-swift: I like this overview - well done!  I hope this meme spreads all over Swift, including our source files.  Super useful to give context for anyone jumping into a new file for the first time.

FindCheck.cmake
woodfell: Yes absolutely. I'll go around inviting everyone as we move through

FindFastCSV.cmake
woodfell: Done. There were a few special cases where it wasn't possible to generalise, but most were ok.

GenericFindDependency.cmake
```diff
@@ -0,0 +1,245 @@
...
+
+  add_library(${x_TargetName} UNKNOWN IMPORTED)
+  set_target_properties(${x_TargetName} PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${x_${x_TargetName}_IncludeDir}")
+  set_property(TARGET ${x_TargetName} APPEND PROPERTY
+      IMPORTED_LOCATION "${x_${x_TargetName}_Library}")
+  message(STATUS "Found ${x_TargetName} from the system at ${x_${x_TargetName}_Library}")
+endfunction()
+
+function(GenericFindDependency)
```
martin-swift: Very thorough, as far as I can tell.   The only thing I might suggest is use some macros to clean up the structure a bit; it's a bit monolithic with a lot of stuff going on.  I see 5-ish major things going on here:

1) collect_arguments()
2) build_location_list()
3) build_source_search_list()
4) build_system_search_name()
5) is_found()

I think that would make it a lot more approachable for people who may not have the same degree of familiarity with the problem.

FindSettings.cmake
```diff
@@ -0,0 +1,5 @@
+include("GenericFindDependency")
+GenericFindDependency(
+    TargetName settings
+    SourceDir "c"
```
benjaminaltieri: this line isn't necessary, as `libsettings` has a top level cmakelists.txt:
```suggestion
```

woodfell: Done

woodfell:
> Placeholder request for ClangTools replacements scripts. Right now all the downstream PRs are defining dummy targets which shouldn't be merged into the trunk.

@benjaminaltieri Definitely. New clang tools support is in the works

GenericFindDependency.cmake
woodfell: Yep good point, it was a bit unwieldy. I've tidied it up a bit, created some helper macros and functions and added a load more comments to explain what's happening where.

woodfell: I've also changed all the parameters to be upper case which is more consistent with cmake style

woodfell:
@benjaminaltieri @jbangelo @martin-swift 
I've finished off the new clang tools modules. I have split them up in to ClangTidy and ClangFormat since not every project uses both. Documentation at the header should be fairly self explanatory. 

I have updated all the other PRs to use these modules and removed all the temporary hack targets. The only project which doesn't use the generic tidy/format targets is starling. It still uses this module but it has custom scripts in the repo. Starling is probably worth revisiting though, the script uses some hard coded flags to clang-tidy which passes, but as soon as it is run against the compile command database exported from cmake it throws loads of errors.

pmiettinen:
Should default SN `.clang-tidy` file be provided? See for example https://github.com/swift-nav/starling/blob/master/.clang-tidy

ClangFormat.cmake
```diff
@@ -0,0 +1,190 @@
...
+# a name clash
+#
+# The parameter SCRIPT can be used to specify a custom formatting command instead
+# of calling clang-format directly. This should be executable and will be called
+# with a single argument of either 'all' or 'diff' according to the target
+#
+# clang-format-all-${PROJECT_NAME} will run `<script> all`
+# clang-format-diff-${PROJECT_NAME} will run `<script> diff`
+#
+# If a script is not specified explicitly this function will first search for an
```
jbangelo: If this is being included just because of the startling repo, I would suggest we remove the added complexity and change how the starling repo works. I added the scripts to starling to simplify our cmake, but using a common cmake implementation accomplished the same goal.

woodfell: While it is true that starling is currently the only project using a custom script I'm not sure that is a reason to get rid of it. Starling won't actually work with the new module, it throws out hundreds of linter errors when running against the compiler_commands.json generated by cmake. But I can imagine this ability being useful in other projects.

Albatross actually has a custom target which calls cppcheck, unfortunately it never actually calls find_program() so it never works. That might be another good candidate to use a custom script.

woodfell:
> Should default SN `.clang-tidy` file be provided? See for example https://github.com/swift-nav/starling/blob/master/.clang-tidy

@jbangelo @martin-swift 
I don't have any strong opinions on this, but I can look in to it. Does anyone else have any thoughts?

benjaminaltieri:
> @jbangelo @martin-swift
> I don't have any strong opinions on this, but I can look in to it. Does anyone else have any thoughts?

let's let individual repos decide what linting they use. it's important that the module supports pulling this in (possibly a custom script as well? i.e. not only `.clang_tidy` file)

ClangFormat.cmake
```diff
@@ -0,0 +1,190 @@
...
+        clang-format34 clang-format-3.4
+        clang-format
+       )
+  endif()
+  find_program(CLANG_FORMAT NAMES ${x_CLANG_FORMAT_NAMES})
+
+  if("${CLANG_FORMAT}" STREQUAL "CLANG_FORMAT-NOTFOUND")
+    # clang-format not found, can't continue
+    message(WARNING "Could not find appropriate clang-format, target disabled")
+    return()
```
benjaminaltieri: can we make this a fatal error?

ClangTidy.cmake
```diff
@@ -0,0 +1,258 @@
...
+  set(argOption "")
+  set(argSingle "")
+  set(argMulti "TARGETS")
+
+  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
+
+  foreach(target ${x_TARGETS})
+    if(NOT TARGET ${target})
+      message(WARNING "clang-tidy enabled for non-existent target ${target}")
+      continue()
```
benjaminaltieri: should this be a fatal error?

ClangTidy.cmake
```diff
@@ -0,0 +1,258 @@
...
+        clang-tidy35 clang-tidy-3.5
+        clang-tidy34 clang-tidy-3.4
+        clang-tidy
+       )
+  endif()
+  find_program(CLANG_TIDY NAMES ${x_CLANG_TIDY_NAMES})
+
+  if("${CLANG_TIDY}" STREQUAL "CLANG_TIDY-NOTFOUND")
+    message(WARNING "Could not find appropriate clang-tidy, target disabled")
+    return()
```
benjaminaltieri: can we make this fatal?

ClangFormat.cmake
```diff
@@ -0,0 +1,190 @@
...
+  if(${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
+    # This is the top level project, ie the CMakeLists.txt which cmake was run
+    # on directly, not a submodule/subproject. We can do some special things now.
+    # The option to enable clang formatting will be enabled by default only for
+    # top level projects. Also the top level project will create an alias target
+    # clang-format-all against the project specific target
+    set(top_level_project ON)
+  endif()
+
+  option(${PROJECT_NAME}_ENABLE_CLANG_FORMAT "Enable auto-formatting of code using clang-format" ${top_level_project})
```
benjaminaltieri: maybe the default for this can be the AND of `${top_level_project}` and another command line specified option `ENABLE_CLANG_FORMAT` that isn't target specific? defaulting that to disabled along with fatal errors on failed setup attempt might help keep CI healthy (i.e. clear errors when something is going wrong in the generate step) while allowing local users to disable generically in the case their env isn't set up right

ClangFormat.cmake
```diff
@@ -0,0 +1,190 @@
...
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+        )
+  endif()
+endfunction()
+
+# External function to create clang-format-* targets. Call according to the
+# documentation in the file header.
+function(swift_setup_clang_format)
+  set(argOption "")
+  set(argSingle "SCRIPT")
```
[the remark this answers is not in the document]
benjaminaltieri: ok this is apparently not possible. dismissing (ref: https://stackoverflow.com/questions/46373858/how-do-i-specify-a-clang-format-file)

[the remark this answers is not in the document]
benjaminaltieri: cc @pmiettinen

ClangFormat.cmake
woodfell: Done. Added global options for both tidy and format targets - SWIFT_ENABLE_CLANG_TIDY and SWIFT_ENABLE_CLANG_FORMAT. Enabled by default, they feed in to the default value for the project specific options.

ClangTidy.cmake
woodfell: Done

ClangTidy.cmake
woodfell: Done

ClangFormat.cmake
woodfell: Done

ClangFormat.cmake
```diff
@@ -0,0 +1,201 @@
...
+  set(argMulti "CLANG_FORMAT_NAMES")
+
+  cmake_parse_arguments(x "${argOptions}" "${argSingle}" "${argMulti}" ${ARGN})
+
+  set(default_option_state ON)
+
+  # Global clang-format enable option, influences the default project specific enable option
+  option(SWIFT_ENABLE_CLANG_FORMAT "Enable auto-formatting of code using clang-format globally" ON)
```
benjaminaltieri: could probably drop the `SWIFT_` part

ClangTidy.cmake
```diff
@@ -0,0 +1,268 @@
...
+  set(argOption "")
+  set(argSingle "SCRIPT")
+  set(argMulti "CLANG_TIDY_NAMES" "TARGETS" "PATTERNS")
+
+  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
+
+  set(default_option_state ON)
+
+  # Global clang-tidy enable option, influences the default project specific enable option
+  option(SWIFT_ENABLE_CLANG_TIDY "Enable auto-linting of code using clang-tidy globally" ON)
```
benjaminaltieri: same here

ClangFormat.cmake
```diff
@@ -0,0 +1,201 @@
...
+        clang-format35 clang-format-3.5
+        clang-format34 clang-format-3.4
+        clang-format
+       )
+  endif()
+  find_program(CLANG_FORMAT NAMES ${x_CLANG_FORMAT_NAMES})
+
+  if("${CLANG_FORMAT}" STREQUAL "CLANG_FORMAT-NOTFOUND")
+    # clang-format not found, can't continue
+    message(WARNING "Could not find appropriate clang-format, target disabled")
```
jbangelo: Do we want to let CMake continue on if we can't find `clang-format`? It would seems more likely that we'd want to fail generation.

I guess the calling project could check for the existence of the target and error out there.

benjaminaltieri: Yeah, Matt and I went back and forth on this (hence the revert), my take was that explicitly disabling the format functions via command line would encourage proper dev env setup but then require all non-linting CI scripts to add those flags to avoid errors. :shrug:

jbangelo: As long as the build fails if it can't properly lint/format the code. I guess that would probably happen if the lint/format target doesn't exist. If the build doesn't fail that's bad.

ClangFormat.cmake
woodfell: Done

ClangTidy.cmake
woodfell: Done

ClangFormat.cmake
woodfell: Addressed comments by adding a `REQUIRED` flag to the setup functions. If specified the clang modules will throw a fatal error if the targets can't be created for any reason, and a warning otherwise. ENABLE_CLANG_FORMAT/TIDY changed to be a global on/off switch rather than feeding the default project options.

PR #3, opened 2019-05-08
[HACKATHON] Create module for code coverage automation with llvm-cov

benjaminaltieri:
Based on https://github.com/StableCoder/cmake-scripts#code-coverage-code-coveragecmake

Tested with clang and llvm-cov (supposed to work with gcc and gcov/lcov too).

More information:
 - http://www.stablecoder.ca/2018/10/30/full-cmake-helper-suite.html
 - https://clang.llvm.org/docs/SourceBasedCodeCoverage.html

### DIFF
--- a/CodeCoverage.cmake
+++ b/CodeCoverage.cmake
@@ -1,0 +1,475 @@
+#
+# Copyright (C) 2018 by George Cave - gcave@stablecoder.ca
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# USAGE: To enable any code coverage instrumentation/targets, the single CMake
+# option of `CODE_COVERAGE` needs to be set to 'ON', either by GUI, ccmake, or
+# on the command line.
+#
+# From this point, there are two primary methods for adding instrumentation to
+# targets: 1 - A blanket instrumentation by calling `add_code_coverage()`, where
+# all targets in that directory and all subdirectories are automatically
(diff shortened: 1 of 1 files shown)


benjaminaltieri:
@woodfell @jbangelo I know underlying modules are still in-flight but happy to get any comments/ideas on this implementation!

benjaminaltieri:
@woodfell considering your consistent treatment of the clang tidy and format modules, should this be revamped to follow those more closely? The underlying procedure of this third party script is relatively similar in that it creates a top level target in addition to sub-targets (though these will be per build target). What it doesn't have is the subproject guards which I think we would want to add. With your assistance/review, @wurmmi and I would likely be able to do the leg work (mob programming candidate!). thoughts?

CodeCoverage.cmake
```diff
@@ -0,0 +1,475 @@
...
+      execute_process(COMMAND ${LLVM_COV_PATH} --version
+                      OUTPUT_VARIABLE LLVM_COV_VERSION_CALL_OUTPUT)
+      string(REGEX MATCH
+                   "[0-9]+\\.[0-9]+\\.[0-9]+"
+                   LLVM_COV_VERSION
+                   ${LLVM_COV_VERSION_CALL_OUTPUT})
+
+      if(LLVM_COV_VERSION VERSION_LESS "7.0.0")
+        message(
+          WARNING
```
benjaminaltieri: should consider turning this into a fatal error

wurmmi:
Once we merge this PR, we need to make sure to update the submodule pointer for libswiftnav. 
This is because https://github.com/swift-nav/libswiftnav/pull/89 was already merged to master, but points to this PR.

CodeCoverage.cmake
wurmmi: Let's keep it a warning for now - we are currently using clang-6.0 in specific targets in Jenkins. Enforcing version >= 7.0.0 here would break compatibility entirely there.
Also, so far I'm not using the `EXCLUDE` option at all.

wurmmi:
The module is already included in https://github.com/swift-nav/libswiftnav/pull/89 and I'm currently working on the integration into starling in https://github.com/swift-nav/starling/pull/2372.
It seems to work fine so far - it's just the clang version check we might consider to revisit once we use the EXCLUDE option in the future.

Thus - LGTM, let's merge.

PR #5, opened 2019-06-17
Specify header file name for non-standard installs

woodfell:
Some package header files can't be found automatically because they are installed to non-standard locations. By default GenericFindDependencies will search for a system header file called "lib${TARGET}/${TARGET}.h", for example the module FindSwiftnav (libswiftnav) will try to find "libswiftnav/swiftnav.h". Some packages install using a different naming scheme, libswiftnav installs "swiftnav/bits.h" (among others). This PR specifies different header file names for those packages so they can be found properly.

### DIFF
--- a/FindGnss-converters.cmake
+++ b/FindGnss-converters.cmake
@@ -2,5 +2,6 @@ include("GenericFindDependency")
 GenericFindDependency(
     TARGET gnss_converters
     SOURCE_DIR "gnss-converters/c"
+    SYSTEM_HEADER_FILE "gnss-converters/nmea.h"
     SYSTEM_INCLUDES
     )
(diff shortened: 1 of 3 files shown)


benjaminaltieri:
does this only affect repos that use `SWIFT_PREFERRED_DEPENDENCY_SOURCE=system`? (piksi_apps, buildroot, etc) Or will all cmake/common users need this fix?

woodfell:
The affected modules were unable to find dependencies from the system because of weirdly named header files. This just fixes the search for these modules, users of these packages will need to be updated with this fix but only if they are using dependencies from the system. Currently this is just piksi_apps/buildroot. Everything else uses dependencies from source and so are unaffected.

PR #8, opened 2019-06-26
Add subproject guard for code coverage all targets

benjaminaltieri:


wurmmi:
LGTM, tested with https://github.com/swift-nav/starling/pull/2372 and seems to work.

PR #9, opened 2019-07-08
Support custom file patterns in clang-format

woodfell:
A minor addition, allows a project to specify a custom set of file matching patterns to pass to `git ls-files` to generate a list of files to format. Required for starling to maintain the behaviour of the old formatting script


benjaminaltieri:
Looks like #10 has the same changes, but igtm

woodfell:
Yup, 2 separate features, but #10 was based off this branch to support another PR

PR #14, opened 2019-07-30
Fixed the clang-tidy-diff target to properly pipe

jbangelo:


ClangTidy.cmake
```diff
@@ -278,7 +278,6 @@ function(swift_setup_clang_tidy)
         ${${PROJECT_NAME}_CLANG_TIDY} ${x_EXTRA_ARGS} -p ${CMAKE_BINARY_DIR} --export-fixes=${CMAKE_CURRENT_SOURCE_DIR}/fixes.yaml
           `git ls-files ${srcs}`
         DIFF_COMMAND
-        git diff --diff-filter=ACMRTUXB --quiet -- ${srcs} ||
```
jbangelo: Not entirely sure if this is the correct change. Maybe this should have been `&&` instead of `||`?

martin-swift: yea, not sure what was going on here, but it definitely looks extraneous to me.

benjaminaltieri: what is the behavior you are seeing? I believe this is meant to circumvent the check if the source code hasn't changed. no idea how it is or isn't working though.

martin-swift: but isn't it redundant?  Like, if nothing has changed you'll end up running clang-tidy on the empty set of files which should be a no-op, no?

jbangelo: I was seeing it not do anything. It was executing `git diff ...` which succeeded and then the `||` operator short circuited and didn't run `clang-tidy`.

When no changes to files are present this PR fails, `clang-tidy` complains about no input files and prints the help output.

I'll figure out an alternative command that'll work in both situations. Probably a use of `xargs`, similar to what the `clang-format` target does.

woodfell: @jbangelo I got rid of xargs for clang-format because it has some unhelpful side effects. If this can wait until I'm back in the office next week I can lend a hand properly, at the moment I'm just on the phone and don't have access to everything.

woodfell: I think the problem here is that the first part of this command (before the `||`) is not passing the branch `master` as an argument. git is therefore looking for uncommited files rather than files which differ to master. The whole point of the `||` operator is to prevent `clang-tidy` being run with an empty list of files so it is weird to hear that you have seen exactly that. I wasn't able to cause it this morning, do you have an example of it happening @jbangelo ? I'm going to push up a fix to this branch which will restore the `||` and add `--name-only master` to the initial `git diff` command

jbangelo: @woodfell Looks like you're right. I've pushed a branch in starling to test this out (jbangelo/intentional-failure). In that branch using the cmake files from `master` `clang-tidy` doesn't show any errors and doesn't appear to run on any files. When I checkout this branch of the cmake files `clang-tidy` does find an error.

woodfell: Good to hear, let's merge

PR #15, opened 2019-07-30
Add format check

benjaminaltieri:
Useful for calling checks from Jenkins instead of needing to stuff bash into the groovy scripting ala: https://github.com/swift-nav/starling/blob/c34b1c2e330d5d4002d80aee72f6af821281d4ce/Jenkinsfile#L557-L565


benjaminaltieri:
> I like it! I wonder if we could do something similar for the clang-tidy check as well?

Tag you're it! 🙌

benjaminaltieri:
for an example failure: https://jenkins.ci.swift-nav.com/blue/organizations/jenkins/swift-nav%2Flibpal/detail/PR-5/13/pipeline

PR #19, opened 2019-08-15
Clear out function args before including dependencies

woodfell:


GenericFindDependency.cmake
```diff
@@ -70,6 +70,10 @@ function(search_dependency_source)
   foreach(P ${x_SOURCE_SEARCH_PATHS})
     if(EXISTS "${P}/CMakeLists.txt")
       message(STATUS "Found ${x_TARGET} source code in ${P}")
+      foreach(i RANGE 0 100)
```
martin-swift: can you throw in a comment briefly describing the situation you discovered yesterday?  Otherwise people are going to have no idea why this is here...

woodfell: Done

woodfell: Maybe not so brief, but it explains what's going on :)

GenericFindDependency.cmake
```diff
@@ -70,6 +70,47 @@ function(search_dependency_source)
   foreach(P ${x_SOURCE_SEARCH_PATHS})
     if(EXISTS "${P}/CMakeLists.txt")
       message(STATUS "Found ${x_TARGET} source code in ${P}")
+      # Function arguments are automatically parsed out in to numbered variables
```
martin-swift: :+1:

PR #23, opened 2019-09-11
Project project_BUILD_* variables

woodfell:
In some places internal cmake variables are being set manually, either on the command line or from a higher level project. Setting `project_BUILD_${feat}` vars breaks the internal feature enable/disable logic and must not be allowed. Projects can TEST `project_BUILD_${feat}` to decide whether to build something or not, for example unit tests, but they must never be set manually. The correct way is to set `project_ENABLE_${feat}` instead.

This does not fix any place which actually does this, just ensure that an error is thrown if someone uses these options erroneously.


benjaminaltieri:
This only projects against setting these variables in the commandline or parent project right? Are there also cases where these are being set within the project itself after Swift options is called?

woodfell:
They could be overwritten later on, but there is nothing we can really do about that. I keep coming across cases of `BUILD_*` being set from a higher level project, or in a Jenkinsfile, that is the behaviour which really needs to be protected against.

PR #28, opened 2019-10-01
Added ability to run code coverage when unit test fails and support t…

RReichert:
There are two major changes introduced here.

1. Originally, any code coverage target whose underlying target ran and returned a status error code of anything other than zero, the code coverage target wouldn't try and produce a report. By introducing the ` || true` entry, we've disregarded the error code and continued along with reporting the test coverage anyways. This would need to be looked and approved by someone with better knowledge of the system then myself as this might break the desired behaviour of the build system.

2. Because in my setup I've got multiple versions of GCC running (open in my systems directory and an a number in `/opt/gcc/*`, I need a way to choose which `gcov` program to use as it needs to be bound to the correct `gcc` version for code coverage to work. This can be done by explicitly passing the `gcov` version to `lcov` command. Existing users shouldn't be affected as cmake will pickup the `gcov` within the system's `$PATH` and pass it explicitly to the `lcov` command, and if open isn't found, it just doesn't inform `lcov`.

Users can work with non system compilers via the following example cmake command:

```
cmake \
  -DCMAKE_PREFIX_PATH:PATH=/opt/gcc/7.4.0 \
  -DCMAKE_C_COMPILER:PATH=/opt/gcc/7.4.0/bin/gcc \
  -DCMAKE_CXX_COMPILER:PATH=/opt/gcc/7.4.0/bin/g++ \
  ...
```

### DIFF
--- a/CodeCoverage.cmake
+++ b/CodeCoverage.cmake
@@ -86,6 +86,7 @@ if (NOT LLVM_COV_PATH)
     endif()
     find_program(LLVM_COV_PATH ${LLVM_COV_NAME})
 endif()
+find_program(GCOV_PATH gcov)
 find_program(LCOV_PATH lcov)
 find_program(GENHTML_PATH genhtml)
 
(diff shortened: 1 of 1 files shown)


benjaminaltieri:
I second the above idea, we should always make sure to test such changes in one or more repos since this is a shared resource and that means extra care for validating behavioral changes is warranted

RReichert:
I agree with you both. To address both your concerns, I've created two pull requests:

* https://github.com/swift-nav/starling/pull/2648 - Passing Unit Tests
* https://github.com/swift-nav/starling/pull/2647 - Failing Unit Test

With both pull requests, I've taken `starling` repository and updated the `cmake/common` sub-module to point to this branch and update the `third_party/gnss-converters` to point to a branch whose common cmake points to this branch (the reason for this is because for whatever reason, whenever I run `ccov-test-starling-runner` it uses the cmake common from that third party library).

The difference between one pull request to the other is that one has a guaranteed failing unit test and another will pass.

I haven't check the results yet for the pull requests, but hopefully it works.

benjaminaltieri:
I might suggest to split these into separate PRs, since the gcov change seems pretty straightforward and useful.

PR #34, opened 2019-10-18
Don't use the name 'starling' in common functions

woodfell:


woodfell:
Heh, I must have given it a very thorough review :)

PR #35, opened 2019-10-21
Import TestTargets helper module from starling

woodfell:


TestTargets.cmake
```diff
@@ -0,0 +1,226 @@
...
+# sense for gtest based test suites and uses the parallel test helper program 
+# from the googletest package. The non-parallel target is always created.
+#
+# The other function, swift_add_test_runner(), can be used to create a do-... 
+# target which pointer at some other command. This can be used to invoke a test 
+# which is actually a shell script or some other executable which is available 
+# but not built from source
+#
+# swift_add_test_runner(test-suite-name
+#    COMMAND <shell-command-to-run-test>
```
RReichert: There are a few things to mention about this option:

1. its not only shell commands that can be run, they can be cmake targets (at which point you don't need to add that as a dependency)
2. if the intention is to use shell commands to run, might want to add the option of WORKING_DIRECTORY as well
3. mention that the command can be followed by arguments that will be passed in

TestTargets.cmake
```diff
@@ -0,0 +1,226 @@
...
+# will disable post build tests from running as part of 'make all'. The post-build 
+# targets will still be created and can be invoked manually.
+#
+
+include(LanguageStandards)
+include(CodeCoverage)
+
+option(AUTORUN_TESTS "Automatically run post-build tests as part of 'all' target" ON)
+if(AUTORUN_TESTS)
+  set(autorun ALL)
```
RReichert: might want mangle that name to be more unique, otherwise it might pollute the global scope since this is normally included into global scope

TestTargets.cmake
RReichert: There are a few things to mention about this option:

1. its not only shell commands that can be run, they can be cmake targets (at which point you don't need to add that as a dependency)
2. if the intention is to use shell commands to run, might want to add the option of WORKING_DIRECTORY as well
3. mention that the command can be followed by arguments that will be passed in

TestTargets.cmake
```diff
@@ -0,0 +1,226 @@
...
+
+function(swift_add_test_runner target)
+  set(argOption "POST_BUILD")
+  set(argSingle "COMMENT")
+  set(argMulti "COMMAND" "DEPENDS")
+
+  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
+
+  if(x_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "swift_add_test unparsed arguments - ${x_UNPARSED_ARGUMENTS}")
```
RReichert: `swift_add_test_runner` rather then `swift_add_test`

TestTargets.cmake
```diff
@@ -0,0 +1,226 @@
...
+      COMMAND ${x_COMMAND}
+      COMMENT "Running ${x_COMMENT}"
+      )
+  add_dependencies(do-all-tests do-${target})
+  if(x_DEPENDS)
+    add_dependencies(do-${target} ${x_DEPENDS})
+  endif()
+
+  if(x_POST_BUILD)
+    add_custom_target(post-build-${target}
```
RReichert: based on your documentation, shouldn't you have something like this somewhere?

```
add_custom_command(ALL POST_BUILD COMMAND ${x_COMMAND})
```

I fail to see how the code here will execute after the ALL target is build.

TestTargets.cmake
```diff
@@ -0,0 +1,226 @@
...
+      do-${target}
+      COMMAND ${target}
+      COMMENT "Running ${x_COMMENT}"
+      )
+  add_dependencies(do-${target} ${target})
+  target_code_coverage(${target} AUTO ALL)
+
+  if(x_PARALLEL)
+    add_custom_target(parallel-${target}
+        COMMAND ${PROJECT_SOURCE_DIR}/third_party/gtest-parallel/gtest-parallel $<TARGET_FILE:${target}>
```
RReichert: I cant remember if this will fail on windows bat script, I believe you need to be explicit about specify the interpreter before hand

TestTargets.cmake
woodfell: I've wrapped this section up in a macro and am now calling it from the 2 setup functions. This stops the variables from leaking out in to the parent scope but also stops the helper targets being created if there aren't any tests actually defined.

TestTargets.cmake
woodfell: Done

TestTargets.cmake
woodfell: It actually gets run by the `do-post-build-tests` target. This is included in `ALL` based on the `AUTORUN_TESTS` option, individual test targets don't need to have that.

TestTargets.cmake
woodfell: Not sure, you may be correct. This parallel option is pretty nasty with its hard coded path and (as far as I know) is only used `starling` which doesn't work on windows.

TestTargets.cmake
woodfell: Clarified the documentation. WORKING_DIRECTORY wasn't actually an option but it is useful so I've added it.

PR #38, opened 2019-10-25
Change FindCheck to look for submodule

woodfell:


mfine:
Thanks for adding libcheck everywhere - I got blocked in gnss converters trying to add a test because of issues with libcheck (despite having it installed locally).

PR #39, opened 2019-10-30
Cleaned up some cached variables after including benchmark

jbangelo:


jbangelo:
It's ugly but I tested this in the starling repo and it appears to solve our problem.

PR #41, opened 2019-12-20
Print out the commit and path for each source submodule that has been included in the build.

martin-swift:
This is for ORI-834; I finally am building the machinery to check for consistency between our multiple copies of submodules.  There will be some scripts in orion that consume this output.
cc: @benjamin0 @mfine

### DIFF
--- a/GenericFindDependency.cmake
+++ b/GenericFindDependency.cmake
@@ -69,7 +69,7 @@ function(search_dependency_source)
 
   foreach(P ${x_SOURCE_SEARCH_PATHS})
     if(EXISTS "${P}/CMakeLists.txt")
-      message(STATUS "******************* Found ${x_TARGET} source code in ${P}")
+      message(STATUS "Found ${x_TARGET} source code in ${P}")
       # Function arguments are automatically parsed out in to numbered variables
       # ARG#, and a complete argument list is stored in ARGN. Variables are passed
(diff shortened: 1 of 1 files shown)


martin-swift:
@woodfell - when this looks good to you, can you approve?  Thanks.

martin-swift:
`*.used` will be output at generation time; presumably a submodule update will trigger a re-gen.  But even if it doesn't, that's probably fine, since Jenkins always makes a fresh clone.

RReichert:
I haven't tried to run the command, so maybe I'm missing something, but won't you have an overwrite of the file if the target has two different submodule commits within the git hierarchy? For instance, with `starling` commit `0133537f2205f5568f546b2c4053d8630cd10701`, `libswiftnav` has two inclusions `127202b2597d5544e8fa182eae0b525c83636d80` and `acdc609354d3917bb80361c5a8487d094ee00c93`.

martin-swift:
But only one will be included, and the whole point of that snippet is to record which one was used.

woodfell:
This is part of `GenericFindDependency` which protects against double inclusion of submodules. Only one can ever be included in a single build. The other half of this work is something which will compare the version which was included in the build against the list of all versions checked out in the source tree and flag up potential issues.

martin-swift:
going to merge now...

RReichert:
Ohhh I see, seems like a good solution.

PR #52, opened 2020-04-22
[POS-51] Update the FindStarling modules

jbangelo:
This updates the `FindStarling` modules to find the new interface library for the PVT engine.

### DIFF
--- a/FindStarling.cmake
+++ b/FindStarling.cmake
@@ -7,3 +7,9 @@ GenericFindDependency(
     SYSTEM_HEADER_FILE "pvt_driver/runner/pvt_runner.h"
     SYSTEM_INCLUDES
 )
+
+mark_target_as_system_includes(sensorfusion)
+mark_target_as_system_includes(pvt_driver)
+mark_target_as_system_includes(pvt-engine)
+mark_target_as_system_includes(pvt-common)
+mark_target_as_system_includes(starling-util)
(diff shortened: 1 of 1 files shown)


FindStarling.cmake
```diff
@@ -2,7 +2,14 @@ include("GenericFindDependency")
...
+    SYSTEM_HEADER_FILE "pvt_driver/runner/pvt_runner.h"
     SYSTEM_INCLUDES
-    )
+)
+
+mark_target_as_system_includes(sensorfusion)
+mark_target_as_system_includes(pvt_driver)
+mark_target_as_system_includes(pvt-engine)
+mark_target_as_system_includes(pvt-common)
+mark_target_as_system_includes(starling-util)
```
RReichert, line 15: should these check to make sure that the targets exists before marking them? `if (TARGET sensorfusion) ... endif()`, on the off chance that we might rename/remove some of these targets in the future.

jbangelo: It looks like `mark_target_as_system_includes` will silently do nothing if the targets don't exist.

Are you thinking that we should be failing or giving a warning if the targets don't exist?

RReichert: I didn't expect that behaviour of the `mark_target_as_system_includes` function (sorry didn't dig to deep into it). I'd probably go with a warning, mostly because cmake is an external library library.

RReichert: my comment is really just a nit. I'll leave matt to approve the request since he knows this portion of the cmake work inside out. I don't see any isssues with this PR, I'm happy to approve.

PR #53, opened 2020-05-06
Add FindSwiftlets.cmake

IsakTjernberg:


FindSwiftlets.cmake
```diff
@@ -0,0 +1,8 @@
+include("GenericFindDependency")
+option(swiftlets_ENABLE_TESTS "" OFF)
+option(pal_ENABLE_EXAMPLES "" OFF)
```
woodfell, line 3: This option is already set in `FindPal.cmake`

FindSwiftlets.cmake
```diff
@@ -0,0 +1,8 @@
...
+GenericFindDependency(
+    TARGET swiftlets
+    SOURCE_DIR swiftlets
```
woodfell, line 6: This isn't strictly speaking required - `GenericFindDependency` searches a load of paths based on the target name to look for the submodule. It won't hurt to leave it there, it's just not required.

IsakTjernberg: I left this here as it seems to be used by quite many other modules also.

woodfell: Yeah, it's really only needed when the source code is in a directory with a different name from the target name, or when the source code is in a subdirectory in the submodule (like gnss-converters or libsbp). In this case the names as the same though

PR #56, opened 2020-05-14
Number of changes were made, too many many to list all at once

RReichert:


Valgrind.cmake
```diff
@@ -153,43 +164,50 @@ macro(_valgrind_executable_setup target)
   endif()
 endmacro()
 
-macro(_valgrind_tools_setup target tool)
-  set(argOption "")
```
RReichert, line 157: Missing trace children which you've specified in the documentation

Valgrind.cmake
```diff
@@ -153,43 +164,50 @@ macro(_valgrind_executable_setup target)
...
+  set(reports_directory ${working_directory}/valgrind-reports)
 endmacro()
 
 function(swift_add_valgrind_memcheck target)
-  set(argOption TRACE_CHILDREN CHILD_SILENT_AFTER_FORK SHOW_REACHABLE TRACK_ORIGINS UNDEF_VALUE_ERRORS)
+  set(argOption CHILD_SILENT_AFTER_FORK SHOW_REACHABLE TRACK_ORIGINS UNDEF_VALUE_ERRORS)
   set(argSingle LEAK_CHECK)
   set(argMulti "")
 
-  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" "${ARGN}")
```
RReichert, line 184: wont need to call cmake_parse_arguments where since `_valgrind_arguments_setup` does that

Valgrind.cmake
```diff
@@ -208,20 +226,14 @@ function(swift_add_valgrind_memcheck target)
   endif()
 
   if (x_LEAK_CHECK)
-    if (${x_LEAK_CHECK} STREQUAL "yes")
-      list(APPEND valgrind_tool_options --leak-check=yes)
-    elseif (${x_LEAK_CHECK} STREQUAL "full")
-      list(APPEND valgrind_tool_options --leak-check=full)
-    elseif (${x_LEAK_CHECK} STREQUAL "summary")
-      list(APPEND valgrind_tool_options --leak-check=summary)
-    endif()
```
RReichert, line 217: Since there is no else statement to say that its an incorrect entry, I've simplified these statements

Valgrind.cmake
```diff
@@ -153,43 +164,50 @@ macro(_valgrind_executable_setup target)
...
-    set(target_name ${x_NAME})
+    set(target_name ${target_name}-${x_NAME})
+    set(report_folder ${report_folder}-${x_NAME})
   endif()
 
   set(working_directory ${CMAKE_CURRENT_BINARY_DIR})
   if (x_WORKING_DIRECTORY)
     set(working_directory ${x_WORKING_DIRECTORY})
   endif()
-  set(valgrind-reports-dir ${working_directory}/${tool}-reports)
```
RReichert, line 176: needed to change the variable name to use `snake_case` over `kebab-case`. `kebab-case` is normally used within swiftnav for target names, to variables its either `snake_case` or `camelCase`.

Valgrind.cmake
```diff
@@ -208,20 +226,14 @@ function(swift_add_valgrind_memcheck target)
...
-      list(APPEND valgrind_tool_options --leak-check=full)
-    elseif (${x_LEAK_CHECK} STREQUAL "summary")
-      list(APPEND valgrind_tool_options --leak-check=summary)
-    endif()
+    list(APPEND valgrind_tool_options "--leak-check=${x_LEAK_CHECK}")
   endif()
 
   add_custom_target(${target_name}
-    COMMENT "Valgrind Memcheck is running for \"${target}\" (output: \"${valgrind-reports-dir}/\")"
-    COMMAND ${CMAKE_COMMAND} -E remove ${valgrind-reports-dir}/${target}*
```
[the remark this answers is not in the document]
Davinco: Yes, the reason was when I ran all five unit tests in libpal with `make do-all-valgrind`, only the last test would be saved since I deleted the folder from the previous tests. So in the previous commit all tests related to a valgrind-tool was saved in a common folder.

Valgrind.cmake
Davinco: I kept trace children in each function because they all had specific logging options connected to it

Valgrind.cmake
Davinco: nice !

Valgrind.cmake
RReichert: ohh I see, they all had the `APPEND valgrind_tool_options --trace-children=yes` in common so I figured I'd move it to the common area. callgrind and massif both have specialization for it, so at least the functionality shouldn't have changed

Valgrind.cmake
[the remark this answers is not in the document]
RReichert: ohhh i see, i was hoping to compartmentalize the data into their own folders to avoid overwriting other files that might be there, hence a clean start every single time. plus the other thing would be that there are a few clashes with naming. If one were to call the `mutex-test` with different program arguments and save them to a common folder, it would be hard to distinguish with PID corresponded to which test.

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
-# `${CMAKE_CURRENT_BINARY_DIR}`.
-# Example: Setting this option for target `suite-2-profiles` to `/tmp`, outputs the profiling
-# results to `/tmp/valgrind-callgrind-reports/suite-2-profiles*`.
-# 
-# TRACE_CHILDREN invokes the Valgrind tools even on spawned children,
-# normally ignores the spawned processes.
+# This would create `valgrind-callgrind-suite-1` and `valgrind-callgrind-suite-2`,
+# each calling the `unit-tests` executable with different program arguments.
 #
-# CHILD_SILENT_AFTER_FORK instructs Valgrind to hide any debugging or logging output for
```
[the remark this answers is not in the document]
RReichert: took care of this, I've now included this in the common arguments

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
+# XTREE_MEMORY produces an execution tree detailing which piece of code is
+# responsible for heap memory usage. Argument `full` gives 6 different
+# measurements, the current number of allocated bytes and blocks (same values as
+# for allocs), the total number of allocated bytes and blocks, the total number
+# of freed bytes and blocks.
 #
 ### MEMCHECK SPECIFIC OPTIONS:
 #
+# CHILD_SILENT_AFTER_FORK instructs Valgrind to hide any debugging or logging
+# output for a child process resulting from a fork call. This can make the
```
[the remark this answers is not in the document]
RReichert: ahhh yes, my bad, I thought it was specific because it was being added to memcheck only, I've since fixed this.

Valgrind.cmake
RReichert: yeah I guess you are right, the behaviour would be the same, I was just hoping to avoid calilng the method twice.

Valgrind.cmake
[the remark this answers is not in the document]
Davinco: Sure! I don't mind keeping it as it is right now. That was just the reason why I changed it to fit my suggestion with a common folder

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
 #   )
 #
-# WORKING_DIRECTORY enables a user to change the output directory from the default folder
-# `${CMAKE_CURRENT_BINARY_DIR}`.
-# Example: Setting this option for target `suite-2-profiles` to `/tmp`, outputs the profiling
-# results to `/tmp/valgrind-callgrind-reports/suite-2-profiles*`.
-# 
-# TRACE_CHILDREN invokes the Valgrind tools even on spawned children,
-# normally ignores the spawned processes.
+# This would create `valgrind-callgrind-suite-1` and `valgrind-callgrind-suite-2`,
```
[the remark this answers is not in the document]
RReichert: done

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
-# Example: a cmake target is called `unit-tests` and is invoked by the callgrind
-# function `swift_add_valgrind_callgrind(unit-tests)`, resulting in the following
-# cmake targets:
+# In this signature, the asterisk is a wildcard for any of the above available
+# tools, so to call on the `massif` tool, use `swift_add_valgrind_massif`.
 #
-#   - unit-tests-valgrind-callgrind
+# These functions create new cmake targets which runs the `target`'s executable
+# binary through Valgrind's tool. If we had a cmake target called `unit-tests`,
+# and invoke the `swift_add_valgrind_callgrind(unit-tests)`, it would result in
```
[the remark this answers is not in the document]
RReichert: Have a look, I believe the changes reflect what was asked, I'm pretty terrible with English in general, so feel free to change whatever you feel is incorrect.

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
+#     [... list of tool specific options ...]
 #   )
 #
-# Call this function to create a new cmake target which runs the executable
-# created by the specified `target` argument through Valgrind's Callgrind tool.
-# Example: a cmake target is called `unit-tests` and is invoked by the callgrind
-# function `swift_add_valgrind_callgrind(unit-tests)`, resulting in the following
-# cmake targets:
+# In this signature, the asterisk is a wildcard for any of the above available
+# tools, so to call on the `massif` tool, use `swift_add_valgrind_massif`.
```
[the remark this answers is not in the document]
RReichert: done

Valgrind.cmake
```diff
@@ -1,139 +1,148 @@
...
 #
-#   - unit-tests-valgrind-callgrind
+# These functions create new cmake targets which runs the `target`'s executable
+# binary through Valgrind's tool. If we had a cmake target called `unit-tests`,
```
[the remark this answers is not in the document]
RReichert: I've made the chages, hopefully it reads better.

Valgrind.cmake
Davinco: Ok! Is it because of the function's overhead? Is it bad practice to call it more than once?

Valgrind.cmake
```diff
@@ -17,12 +17,11 @@
...
 # In this signature, the asterisk is a wildcard for any of the above available
-# tools, so to call on the `massif` tool, use `swift_add_valgrind_massif`.
+# tools, so to call on the `callgrind` tool, use `swift_add_valgrind_callgrind`.
 #
 # These functions create new cmake targets which runs the `target`'s executable
-# binary through Valgrind's tool. If we had a cmake target called `unit-tests`,
-# and invoke the `swift_add_valgrind_callgrind(unit-tests)`, it would result in
-# the following new cmake targets:
+# binary through Valgrind's tool. Invoking the `swift_add_valgrind_callgrind(unit-tests)`,
+# will result in the following new cmake targets:
```
Davinco, line 24:
```suggestion
# results in the following new cmake targets:
```

Valgrind.cmake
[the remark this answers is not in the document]
Davinco: You are probably better than me anyways :P

Valgrind.cmake
```diff
@@ -1,139 +1,147 @@
...
-# Example: a cmake target is called `unit-tests` and is invoked by the callgrind
-# function `swift_add_valgrind_callgrind(unit-tests)`, resulting in the following
-# cmake targets:
+# In this signature, the asterisk is a wildcard for any of the above available
+# tools, so to call on the `callgrind` tool, use `swift_add_valgrind_callgrind`.
 #
-#   - unit-tests-valgrind-callgrind
+# These functions create new cmake targets which runs the `target`'s executable
+# binary through Valgrind's tool. Invoking the `swift_add_valgrind_callgrind(unit-tests)`,
+# will result in the following new cmake targets:
```
Davinco, line 24:
```suggestion
# results in the following new cmake targets:
```

RReichert: done

Valgrind.cmake
```diff
@@ -153,47 +161,57 @@ macro(_valgrind_executable_setup target)
...
 
-  unset(valgrind_tool_options)
-  if (x_TRACE_CHILDREN)
-    list(APPEND valgrind_tool_options --trace-children=yes)
-  endif()
-
-  if (x_CHILD_SILENT_AFTER_FORK)
-    list(APPEND valgrind_tool_options --child-silent-after-fork=yes)
-  endif()
+  list(APPEND valgrind_tool_options --xml=yes --xml-file=report.xml)
```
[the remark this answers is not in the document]
RReichert: done

Valgrind.cmake
RReichert: I guess it was more me being nit picky. There is only one scenario where I can think there might be a problem if we later on add an option in the tool which accepts multiple arguments. For example, if the option called `IGNORE_FILES`, and the user invoked the function as such:
```
swift_add_valgrind_*(target
  PROGRAM_ARGS
    --gtest_filter=UnitTest.*
  IGNORE_FILES
    src/util.h
    src/memory.cc
)
```
The parser wouldn't know that `IGNORE_FILES` isn't one of the `PROGRAM_ARGS` entries. But that won't be a problem now because we aren't in that scenario, but it would be better to plan for future upgrades to the cmake file.

benjaminaltieri:
Looks like a lot of good conversation here @RReichert @Davinco 

This was shared to me as an example of some valgrind scripting that was implemented a while back for Orion. https://github.com/swift-nav/orion/blob/master/profile.sh

In case it has any interesting usage you may want to steal, thought I'd post it here

RReichert:
> Looks like a lot of good conversation here @RReichert @Davinco
> 
> This was shared to me as an example of some valgrind scripting that was implemented a while back for Orion. https://github.com/swift-nav/orion/blob/master/profile.sh
> 
> In case it has any interesting usage you may want to steal, thought I'd post it here

Had a look, pretty much on par with what we are doing, they've used callgrind and memcheck as their tools. At least we are on the same track.

PR #57, opened 2020-05-15
Profiling integration

Davinco:
Included is the profiling applications Valgrind and Gprof. For Valgrind, three tools are available: memcheck, callgrind and massif.


benjaminaltieri:
One thing, when attempting to view the log, I get this error:
```
Gtk-Message: 17:23:29.678: Failed to load module "canberra-gtk-module"
valkyrie: Parse error [build/test/integration/valgrind-reports/memcheck-SwiftletsIntegration.pvt_sanity_1.yaml/test-swiftlets-integration.log.21995] : You do not have executable permission
valkyrie: Use --help for more information.
```
Which seems to be due to the log having only read permissions. Have you run into this? do we have the outputs being saved in a way that would make them read only?

PR #59, opened 2020-05-26
ESD-1836: added additional massif options

RReichert:
Not much going on in here, except just adding some extra valgrind-massif options.

### DIFF
--- a/Valgrind.cmake
+++ b/Valgrind.cmake
@@ -109,6 +109,18 @@
 #
 ### MASSIF SPECIFIC OPTIONS:
 #
+# DEPTH maximum depth of the allocation trees recorded for detailed snapshots.
+#
+# DETAILED_FREQUENCY frequency of detailed snapshots. With value of `1`,
+# every snapshot is detailed.
+#
+# MAX_SNAPSHOTS the maximum number of snapshots recorded.
+#
+# PEAK_INACCURACY does not necessarily record the actual global memory
+# allocation peak; by default it records a peak only when the global memory
+# allocation size exceeds the previous peak by at least 1.0%. Setting this value
+# to `0.0` gives the true value of the peak.
+#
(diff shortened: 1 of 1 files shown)


Valgrind.cmake
```diff
@@ -109,6 +109,18 @@
 #
 ### MASSIF SPECIFIC OPTIONS:
 #
+# DEPTH maximum depth of the allocation trees recorded for detailed snapshots.
+#
```
Davinco, line 113: Could you also add a unit or range to use as argument to depth?

Valgrind.cmake
```diff
@@ -109,6 +109,18 @@
...
+# DETAILED_FREQUENCY frequency of detailed snapshots. With value of `1`,
+# every snapshot is detailed.
+#
```
Davinco, line 116: Same comment as above, can I use decimal numbers as argument?

Valgrind.cmake
RReichert: I did mention that the parameter expects a number and specified the default value. I've made this change to all the other single variable parameters as well.

There is no upper range within Valgrinds documentation (in my implementation it caps at 200), so I didn't want to specify the range as that isn't set in stone from the looks of things.

Valgrind.cmake
RReichert: :heavy_check_mark:, marked it as `float`, the notation that is used in Valgrind of `m.n` seems a bit strange

RReichert: If you prefer `decimal` over `float` I can use `decimal`

Davinco: No, I prefer float :P looks good :+1:

Valgrind.cmake
Davinco: Super! :+1:

PR #61, opened 2020-06-24
David/profiling tools

Davinco:
Adding the current state of my work to get some feedback.

I've tested Bloaty, Heaptrack and Stackusage in swiftlets-repo doing the following:
* inserting `include(Bloaty)`, `include(Heaptrack)`, `include(Stackusage)` in swiftlets' `CMakeLists.txt`.
* inserting the the following in src/starling_debug's `CMakeLists.txt`:
`swift_add_bloaty(starling-debug-binary)` 
`swift_add_heaptrack(starling-debug-binary PROGRAM_ARGS --config /home/davidenberg/Desktop/replays/starling_debug_test.yaml)`
`swift_add_stackusage(starling-debug-binary PROGRAM_ARGS --config /home/davidenberg/Desktop/replays/starling_debug_test.yaml)`

And then running cmake .. which will take a bit longer to finish due to the fetchcontent-function
And finally `make bloaty-starling-debug-binary heaptrack-starling-debug-binary stackusage-starling-debug-binary -j8`

### DIFF
--- a/Bloaty.cmake
+++ b/Bloaty.cmake
@@ -30,9 +30,9 @@
 #       both - Default, sorts by max(vm, file).
 #
 # WORKING_DIRECTORY
-# This variable enables a user to change the output directory for the tools
-# from the default folder `${CMAKE_CURRENT_BINARY_DIR}`. Setting this option for
-# target `starling-debug-binary` to `/tmp`, outputs the results into
+# This variable changes the output directory for the tool
+# from the default folder `${CMAKE_CURRENT_BINARY_DIR}` to the given argument.
+# Example, using argument `/tmp`, outputs the results to
 # `/tmp/bloaty-reports/
 #
 # NOTE
@@ -131,12 +131,23 @@ function(swift_add_bloaty target)
     list(APPEND resource_options -s ${x_SORT})
   endif()
 
-  add_custom_target(${target_name}
-    COMMENT "bloaty is running on ${target}\ (output: \"${reports_directory}/${output_file}\")"
-    COMMAND ${CMAKE_COMMAND} -E remove_directory ${reports_directory}
-    COMMAND ${CMAKE_COMMAND} -E make_directory ${reports_directory}
-    COMMAND ${CMAKE_COMMAND} -E chdir ${reports_directory} echo \"bloaty with options: ${resource_options}\" > ${reports_directory}/${output_file}
-    COMMAND ${CMAKE_COMMAND} -E env $<TARGET_FILE:bloaty> ${resource_options} $<TARGET_FILE:${target}> >> ${reports_directory}/${output_file}
-    DEPENDS ${target}
-  )
+  if (NOT Bloaty_FOUND)
+    add_custom_target(${target_name}
+      COMMENT "bloaty is running on ${target}\ (output: \"${reports_directory}/${output_file}\")"
+      COMMAND ${CMAKE_COMMAND} -E remove_directory ${reports_directory}
+      COMMAND ${CMAKE_COMMAND} -E make_directory ${reports_directory}
+      COMMAND ${CMAKE_COMMAND} -E chdir ${reports_directory} echo \"bloaty with options: ${resource_options}\" > ${reports_directory}/${output_file}
+      COMMAND ${CMAKE_COMMAND} -E env $<TARGET_FILE:bloaty> ${resource_options} $<TARGET_FILE:${target}> >> ${reports_directory}/${output_file}
+      DEPENDS ${target}
+    )
+  else()
+    add_custom_target(${target_name}
+      COMMENT "bloaty is running on ${target}\ (output: \"${reports_directory}/${output_file}\")"
+      COMMAND ${CMAKE_COMMAND} -E remove_directory ${reports_directory}
+      COMMAND ${CMAKE_COMMAND} -E make_directory ${reports_directory}
+      COMMAND ${CMAKE_COMMAND} -E chdir ${reports_directory} echo \"bloaty with options: ${resource_options}\" > ${reports_directory}/${output_file}
+      COMMAND ${CMAKE_COMMAND} -E env ${Bloaty_EXECUTABLE} ${resource_options} $<TARGET_FILE:${target}> >> ${reports_directory}/${output_file}
+      DEPENDS ${target}
+    )
+  endif()
 endfunction()
--- a/Bloaty.cmake
+++ b/Bloaty.cmake
@@ -39,19 +39,16 @@
 
 cmake_minimum_required(VERSION 3.11.0)
 
-set(resource_name bloaty)
-set(github_repo https://github.com/google/bloaty.git)
-
 include(FetchContent)
 FetchContent_Declare(
-  ${resource_name}
-  GIT_REPOSITORY ${github_repo}
+  bloaty
+  GIT_REPOSITORY https://github.com/google/bloaty.git
   GIT_TAG        origin/master
 )
-FetchContent_GetProperties(${resource_name})
-if(NOT ${resource_name}_POPULATED)
-  FetchContent_Populate(${resource_name})
-  add_subdirectory(${${resource_name}_SOURCE_DIR} ${${resource_name}_BINARY_DIR})
+FetchContent_GetProperties(bloaty)
+if(NOT bloaty_POPULATED)
+  FetchContent_Populate(bloaty)
+  add_subdirectory(${bloaty_SOURCE_DIR} ${bloaty_BINARY_DIR})
 endif()
 
 macro(eval_target target)
@@ -61,7 +58,7 @@ macro(eval_target target)
 
   get_target_property(target_type ${target} TYPE)
   if (NOT target_type STREQUAL EXECUTABLE)
-    message(FATAL_ERROR "Specified target \"${target}\" must be an executable type to register for profiling with \"${resource_name }\"")
+    message(FATAL_ERROR "Specified target \"${target}\" must be an executable type to register for profiling with bloaty")
   endif()
 
   if (NOT ${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
@@ -69,7 +66,7 @@ macro(eval_target target)
   endif()
 endmacro()
 
-function(swift_add_${resource_name} target)
+function(swift_add_bloaty target)
   eval_target(${target})
   
   set(argOption SEGMENTS SECTIONS SYMBOLS COMPILEUNITS)
@@ -82,12 +79,12 @@ function(swift_add_${resource_name} target)
     message(FATAL_ERROR "Unparsed arguments ${x_UNPARSED_ARGUMENTS}")
   endif()
 
-  set(target_name ${resource_name}-${target})
+  set(target_name bloaty-${target})
   set(working_directory ${CMAKE_CURRENT_BINARY_DIR})
   if (x_WORKING_DIRECTORY)
     set(working_directory ${x_WORKING_DIRECTORY})
   endif()
-  set(reports_directory ${working_directory}/${resource_name}-reports)
+  set(reports_directory ${working_directory}/bloaty-reports)
   set(output_file ${target_name}.txt)
 
   unset(resource_options)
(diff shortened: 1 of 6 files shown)


Bloaty.cmake
```diff
@@ -0,0 +1,124 @@
...
+FetchContent_Declare(
+  bloaty
+  GIT_REPOSITORY https://github.com/google/bloaty.git
+  GIT_TAG        origin/master
+)
+FetchContent_GetProperties(bloaty)
+if(NOT bloaty_POPULATED)
+  FetchContent_Populate(bloaty)
+  add_subdirectory(${bloaty_SOURCE_DIR} ${bloaty_BINARY_DIR})
+endif()
```
benjaminaltieri, line 52: while this may be useful on a developer's machine, it would be best to allow the system to provide the tool by default and only enable building if requested. a cmake variable could be defaulted to OFF, and turned ON by a user if desired to manage this

RReichert: > can you add a PR in `swiftlets` with the changes you describe in the PR description as well?

The newly generate image for swift-build is `571934480752.dkr.ecr.us-west-2.amazonaws.com/swift-build:2020-06-26`, if you update the `swiftlets` `Dockerfile` with this new image, the profiling tools should be accessible from anywhere.

Davinco: Updated the cmake modules to first try to find the programs and if not found fetch them from source

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+
+  FetchContent_Declare(
+    bloaty
+    GIT_REPOSITORY https://github.com/google/bloaty.git
+    GIT_TAG        origin/master
+  )
+  FetchContent_MakeAvailable(bloaty)
+endif()
+
+macro(eval_target target)
```
RReichert, line 68: There is a subtle bug here. the source of the problem lies in the fact that `Bloaty.cmake`, `Heaptrack.cmake`, and `Stackusage` all declare the same function `eval_target`. 

Cmake has a number of quarks, one of which is that whenever you call `include(...)`, everything that is defined in the `*.cmake` file becomes accessible to to the caller's scope and all its children. This is quite useful feature to have, the downside is that there is no feature in the language to limit what gets "published", so all variables/function names need to be unique.

Would suggest you change the name to something more specific.

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+
+  if (NOT ${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
+    return()
+  endif()
+endmacro()
+
+function(swift_add_bloaty target)
+  if (NOT ${PROJECT_NAME}_ENABLE_MEMORY_PROFILING)
+    return()
+  endif()
```
RReichert, line 86: I would put this statement under `eval_target`, reason is that if a developer writes code incorrect code and doesn't have the `*_ENABLE_MEMORY_PROFILING` flag turned on, it won't ever check to make sure that target exists and will only be flagged as an error the moment another developer turns the flag on.

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+  if (x_SORT)
+    list(APPEND resource_options -s ${x_SORT})
+  endif()
+
+  add_custom_target(${target_name}
+    COMMENT "bloaty is running on ${target}\ (output: \"${reports_directory}/${output_file}\")"
+    COMMAND ${CMAKE_COMMAND} -E remove_directory ${reports_directory}
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${reports_directory}
+    COMMAND ${CMAKE_COMMAND} -E chdir ${reports_directory} echo \"bloaty with options: ${resource_options}\" > ${reports_directory}/${output_file}
+    COMMAND ${CMAKE_COMMAND} -E env $<TARGET_FILE:bloaty> ${resource_options} $<TARGET_FILE:${target}> ${x_PROGRAM_ARGS} >> ${reports_directory}/${output_file}
```
RReichert, line 139: does this work when not using `FetchContent*`?

Heaptrack.cmake
```diff
@@ -0,0 +1,117 @@
...
+#
+# will explicitly enable these targets from the command line at configure time
+#
+
+find_package(Heaptrack)
+
+if (NOT Heaptrack_FOUND AND ${PROJECT_NAME}_ENABLE_MEMORY_PROFILING)
+  message(STATUS "Heaptrack is not installed on system, will fetch content from source")
+
+  cmake_minimum_required(VERSION 3.11.0)
```
RReichert, line 41: make sure the `cmake_minimum_required` version is consistent across the various `*.cmake* files

Heaptrack.cmake
```diff
@@ -0,0 +1,117 @@
...
+  set(working_directory ${CMAKE_CURRENT_BINARY_DIR})
+  if (x_WORKING_DIRECTORY)
+    set(working_directory ${x_WORKING_DIRECTORY})
+  endif()
+  set(reports_directory ${working_directory}/heaptrack-reports)
+
+  if (NOT Heaptrack_FOUND)
+    add_custom_target(${target_name}
+      COMMENT "heaptrack is running on ${target}\ (output: \"${reports_directory}\")"
+      COMMAND $(MAKE)
```
RReichert, line 101: what is the purpose of this command?

Bloaty.cmake
RReichert: specifically `$<TARGET_FILE:bloat>`

Heaptrack.cmake
```diff
@@ -0,0 +1,117 @@
...
+  if (NOT TARGET ${target})
+    message(FATAL_ERROR "Specified target \"${target}\" does not exist")
+  endif()
+
+  get_target_property(target_type ${target} TYPE)
+  if (NOT target_type STREQUAL EXECUTABLE)
+    message(FATAL_ERROR "Specified target \"${target}\" must be an executable type to register for profiling with heaptrack")
+  endif()
+
+  if (NOT ${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
```
RReichert, line 69: you will need a check to disable this when cross compiling. in `Valgrind.cmake` there is a block:

```
if (CMAKE_CROSSCOMPILING)
  return()
endif()
```

Which specifically disables the ability to run these valgrind tests when cross compiling to a different platform. This will be true for both `Heaptrack` and `Stackusage`, but not for `Bloaty` since bloaty just looks at an ELF file.

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+#
+
+option(${PROJECT_NAME}_ENABLE_MEMORY_PROFILING "Builds targets with memory profiling" OFF)
+
+find_package(Bloaty)
+
+if (NOT Bloaty_FOUND AND ${PROJECT_NAME}_ENABLE_MEMORY_PROFILING)
+  message(STATUS "Bloaty is not installed on system, will fetch content from source")
+
+  cmake_minimum_required(VERSION 3.14.0)
```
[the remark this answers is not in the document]
Davinco: Yes, I'm aware of that. Using your new docker image cmake is of version 3.13.4, but since the programs are preinstalled this is only an issue using the "developer setup"

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+#       vm   - Sort the vm size column from largest to smallest and tells you
+#              how much space the binary will take when loaded into memory.
+#       file - Sort the file size column from largest to smallest and tells you
+#              how much space the binary is taking on disk.
+#       both - Default, sorts by max(vm, file).
+#
+# WORKING_DIRECTORY
+# This variable enables a user to change the output directory for the tools
+# from the default folder `${CMAKE_CURRENT_BINARY_DIR}`. Setting this option for
+# target `starling-binary` to `/tmp`, outputs the results into
```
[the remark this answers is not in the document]
Davinco: This is just an example, but I can change to another target name

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+if (NOT Bloaty_FOUND AND ${PROJECT_NAME}_ENABLE_MEMORY_PROFILING)
+  message(STATUS "Bloaty is not installed on system, will fetch content from source")
+
+  cmake_minimum_required(VERSION 3.14.0)
+  include(FetchContent)
+
+  FetchContent_Declare(
+    bloaty
+    GIT_REPOSITORY https://github.com/google/bloaty.git
+    GIT_TAG        origin/master
```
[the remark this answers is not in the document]
Davinco: I don't necessarily agree with you here. Of course, bugs could be introduced but I think that we should use the latest available version. But I can change to use the latest release in the repo

Bloaty.cmake
```diff
@@ -0,0 +1,142 @@
...
+#       file - Sort the file size column from largest to smallest and tells you
+#              how much space the binary is taking on disk.
+#       both - Default, sorts by max(vm, file).
+#
+# WORKING_DIRECTORY
+# This variable enables a user to change the output directory for the tools
+# from the default folder `${CMAKE_CURRENT_BINARY_DIR}`. Setting this option for
+# target `starling-binary` to `/tmp`, outputs the results into
+# `/tmp/bloaty-reports/
+#
```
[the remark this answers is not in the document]
Davinco: removed for bloaty, added for heaptrack and stackusage

Bloaty.cmake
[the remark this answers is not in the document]
Davinco: changed to starling-debug-binary

Bloaty.cmake
Davinco: changed to specific names

Bloaty.cmake
Davinco: yes it works using both setups

Bloaty.cmake
Davinco: moved to eval macro

Heaptrack.cmake
Davinco: I want to use the most recent cmake functionality, with the fetchcontent_makeavailble() function (available from v3.14) However, I need to disable tests for heaptrack and thus need to use the more manual setup from v3.11

Heaptrack.cmake
Davinco: The program needs to be built before I can use it on a target

Heaptrack.cmake
Davinco: updated heaptrack and stackusage

Bloaty.cmake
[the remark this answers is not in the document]
RReichert: might want to remove the bit that says:

```
#  swift_add_bloaty(<target>
#     [OPTIONS]
#     [WORKING_DIRECTORY working_directory]
#     [PROGRAM_ARGS arg1 arg2 ...]   << this portion is incorrect
#   )
```

Bloaty.cmake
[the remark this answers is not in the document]
RReichert: what I was trying to get at was that the target itself is irrelevant to the result. the comment was more of a nit really

Bloaty.cmake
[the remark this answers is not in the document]
RReichert: had a look, ubuntu `20.04` and macos, they all have 3.14+, so I guess that should be ok.

Bloaty.cmake
Davinco: @RReichert @benjaminaltieri here is the link to the PR in swiftlets: https://github.com/swift-nav/swiftlets/pull/616

benjaminaltieri:
@Davinco I just tried to update the swiftlets PR to use gtest with filtering similar to how the profiling does it (memcheck massif etc) but I can't use the target name that would be used to run the test and add a custom name. Can we add a `NAME` option similar to the profiling/valgrind cmake options? That would allow us to use that pattern which is really useful

benjaminaltieri:
@RReichert can you update your review as well based on the updates that have been made? Let's try to merge in the next day or so

Bloaty.cmake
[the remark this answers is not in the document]
RReichert: also just so you know, cmake repository is publicly visible in github (ie. you don't need to login to github), it its best to avoid exposing internal terminologies to the public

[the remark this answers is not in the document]
Davinco: Good input, I'll keep that in mind :+1:

Davinco:
I've added two new commits:
1. option NAME for stackusage and heaptrack
2. option NUM for bloaty to choose how many rows to display for each level. 

I then rebased on master and force pushed to this branch, that's why all commits is shown.

Bloaty.cmake
```diff
@@ -0,0 +1,168 @@
...
+    list(APPEND resource_options symbols)
+  endif()
+
+  if (x_COMPILEUNITS)
+    list(APPEND resource_options compileunits)
+  endif()
+
+  if (DEFINED resource_options)
+    string(REPLACE ";" "," resource_options "${resource_options}")
+    list(PREPEND resource_options -d)
```
RReichert, line 133: rather than using `PREPEND` which only is available with 3.15, suggest you try out:

```
set(resource_options -d ${resource_options})
```

PR #64, opened 2020-08-27
Update profiling tools

Davinco:
Small update where the major parts are:
* a common cmake target called `do-all-memory-profiling` (added into bloaty, stackusage and heaptrack)
* move profiling results from valgrind into common folder `build/profiling/`
* added debug flag -g when running valgrind memcheck, so it's not dependent on cmake build type.
* formatting


Bloaty.cmake
```diff
@@ -41,21 +41,21 @@
...
 # Running
 #
-# cmake -D<project>_ENABLE_MEMORY_PROFILING=ON ..
+# cmake -D<project>_ENABLE_PROFILING=ON ..
 #
 # will explicitly enable these targets from the command line at configure time
 #
 
-option(${PROJECT_NAME}_ENABLE_MEMORY_PROFILING "Builds targets with memory profiling" OFF)
+option(${PROJECT_NAME}_ENABLE_PROFILING "Builds targets profiling applied" OFF)
```
RReichert, line 54: nit: different comment to the rest

Valgrind.cmake
```diff
@@ -143,38 +148,46 @@
...
+# with the name ${PROJECT_NAME}_ENABLE_PROFILING.
+#
+# Running
+#
+# cmake -D<project>_ENABLE_PROFILING=ON ..
+#
+# will explicitly enable these targets from the command line at configure time
+#
+
+option(${PROJECT_NAME}_ENABLE_PROFILING "Build targets with profiling applied" OFF)
```
RReichert, line 164: nit: different comment to the others

Valgrind.cmake
```diff
@@ -42,7 +43,8 @@
 # `swift_add_valgrind_*` functions:
 #
 #   * NAME
-#   * WORKSPACE
+#   * WORKING_DIRECTORY
```
RReichert, line 46: thanks for fixing that

RReichert (commented):
If you'd like to change MEMORY_PROFILING to just PROFILING, would you mind also updating `CodeProfiling.cmake` to follow similar approach (ie. `CODE_PROFILING` to `PROFILING`)?

Valgrind.cmake
```diff
@@ -203,10 +216,11 @@ macro(_valgrind_arguments_setup _target _tool_name _toolOptions _toolSingle _too
     set(report_folder ${_tool_name}-${x_NAME})
   endif()
 
-  set(working_directory ${CMAKE_CURRENT_BINARY_DIR})
+  set(working_directory ${CMAKE_BINARY_DIR}/profiling)
```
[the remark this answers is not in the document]
Davinco: Ok, but then we should change the general structure of `working_directory` and how it is used.

My interpretation of a working directory is the file path where a program reads from and/or saves files into. Atm, a user cannot directly control the file path where a binary gets executed from because the `working_directory` sets the root dir of `report_directory` and then in `add_custom_target` a `reports_directory/report_folder` gets created, which is the file path where the program gets executed. So if a user sets the variable `working_directory` to `/tmp` (because he wants the program to utilize some files from that location), he still needs to create `/tmp/valgrind-reports/"report-folder"/` in order to get the correct file path. This means that we are basically using working_directory as a variable to control the output directory atm.

To change, I see the following tasks:
* Leave the variable `working_directory` as it has previously been assigned in Valgrind. Update in bloaty, heaptrack and stackusage to be defaulted to `CMAKE_CURRENT_BINARY_DIR` but can be changed via the same variable.
* Create a new variable called `report_directory` that is defaulted to `CMAKE_BINARY_DIR/profiling` but which can be changed and use that in all profiling modules to set the output file path.
* Update the `add_custom_target` so that the binary is actually executed in `working_directory` (no chdir) and then save the output in the `report_directory` file path.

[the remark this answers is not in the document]
RReichert: You are correct in your interpretation, the documentation reflects that. I should have worded my comment differently to reflect this as a feature request since you were going to update the report folder variable. I'm hoping to split the concept of a working directory vs report folder so that the working directory is the folder in which you launch the application on (aka "Present Working Directory") and report folder is where the valgrind output gets dumped to. This request isn't essential, you can ignore it if it takes too long to make the changes.

Davinco:
@RReichert, I've updated all profiling tools to support the variables working_directory and report_directory.

Also renamed code_profiling -> gprof, but I've kept the cmake variable CODE_PROFILING because of the -pg flags that get added to all libraries in the function target_code_profiling, so that those libraries don't get "polluted" when running the other profiling tools

Last, it was quite tricky to get the post_command to work in gprof, when generating a report, and it could probably be updated again but I need some input there.

Please comment a lot :)

benjaminaltieri (approved):
LGTM!

RReichert (approved):

PR #65, opened 2020-09-10
Starling third party fix

RReichert:
## Background

There were two issues during the restructuring of our code base.

The first change that needed to be done was fix the `create_source_search_paths` function. After adding the `starling` folder in the root directory, `FindStarling.cmake` script's usage of `SOURCE_DIR` meant that the `${CMAKE_CURRENT_SOURCE_DIR}/${x_SOURCE_DIR}` folder was choose as the folder for which it would call `add_subdirectory` on to introduce the starling repo's cmake targets.

The second issue was simply that some starling libraries weren't being flagged as exportable by the subdirectory, which meant that if we were to use `-DTHIRD_PARTY_INCLUDES_AS_SYSTEM=false` cmake cache variable, it would fail static analysis as `clang-tidy` would complain about eigen values.


GenericFindDependency.cmake
```diff
@@ -259,9 +263,11 @@ macro(create_source_search_paths)
     list(APPEND x_SOURCE_SEARCH_PATHS "${PROJECT_SOURCE_DIR}/${x_SOURCE_DIR}")
```
woodfell, line 263: I think the real issue here is that this line shouldn't be here, neither should the one 2 lines above. A long time ago we agreed that all submodules will live under `third_party`, these modules shouldn't be searching any part that doesn't have `third_party` in it

woodfell (commented):
One request to fix faulty search paths, other than that looks good

FindNanopb.cmake
```diff
@@ -2,7 +2,7 @@ include("GenericFindDependency")
 option(nanopb_BUILD_GENERATOR "" OFF)
 GenericFindDependency(
   TARGET protobuf-nanopb
-  SOURCE_DIR "third_party/nanopb"
+  SOURCE_DIR "nanopb"
```
RReichert, line 5: this was needed to be updated since one of the deleted paths was working with the original `third_party/nanopd`.

GenericFindDependency.cmake
RReichert: Seems like a good idea to remove the old system. I've since rolled back the changes and used the suggested approach.

RReichert:
**NOTE** these changes are currently being used by the https://github.com/swift-nav/swiftlets/pull/880 PR.

woodfell (approved):
LGTM!

PR #66, opened 2020-09-11
Change implementation of working directory

Davinco:
For some reason I can't use `WORKING_DIRECTORY` with `add_custom_target` for heaptrack, so the change is:
* change dir to working directory and execute heaptrack
* move result file from working dir to report dir using `add_custom_command` and post build


IsakTjernberg (approved):
LGTM

PR #67, opened 2020-09-17
Add script to convert valgrind xml to JUnit xml

Davinco:
Since I'm waiting for other submodules to merge into swiftlets before I can add memcheck tests into jenkins, I tried to move the python scripts from the swiftlets repo to be included in the cmake module and this is result. The swiftlets PR has been updated using this approach.

Now it should be easier to integrate memcheck reporting in jenkins to other repos as well.


scripts/xml2junit_converter.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+#
+# This script converts an xml file into JUnit xml format. JUnit is a testing
```
benjaminaltieri, line 3: is this only for memcheck xml results? it would help to have that mentioned so a future maintainer doesn't try to reuse for an unsupported format

benjaminaltieri: maybe the file should be named `memcheck_xml2junit_converter.py` for clarity

Valgrind.cmake
```diff
@@ -299,6 +308,18 @@ function(swift_add_valgrind_memcheck target)
...
+
+  if (x_GENERATE_JUNIT_REPORT)
+    set(xml_dir ${report_directory}/junit-xml)
+    set(options --${x_JUNIT_OPTIONS})
+    add_custom_command(TARGET ${target_name} POST_BUILD
+      COMMAND ${CMAKE_COMMAND} -E make_directory ${xml_dir}
+      COMMAND ${CMAKE_COMMAND} -Dinput_directory=${report_directory}/${report_folder} 
+                               -Doutput_directory=${xml_dir}
+                               -Djunit_options=${options} 
+                               -P ${CMAKE_SOURCE_DIR}/cmake/common/JunitGenerator.cmake
```
RReichert, line 320: I believe this will fail on some repositories, using `${CMAKE_CURRENT_SOURCE_DIR}` mean you can replace `${CMAKE_SOURCE_DIR}/cmake/common/JunitGenerator.cmake` with `${CMAKE_CURRENT_SOURCE_DIR}/JunitGenerator.cmake`.

Valgrind.cmake
```diff
@@ -266,8 +275,8 @@ macro(setup_custom_target valgrind_tool target_name)
 endmacro()
 
 function(swift_add_valgrind_memcheck target)
-  set(argOption SHOW_REACHABLE TRACK_ORIGINS UNDEF_VALUE_ERRORS)
-  set(argSingle LEAK_CHECK)
+  set(argOption SHOW_REACHABLE TRACK_ORIGINS UNDEF_VALUE_ERRORS GENERATE_JUNIT_REPORT)
+  set(argSingle LEAK_CHECK JUNIT_OPTIONS)
```
RReichert, line 279: Wouldn't this be more appropriate in the `argMult` section since its possible in the future to add multiple options to the script? Also would be useful to explain how to use the skip test option, ex what is the format of it.

JunitGenerator.cmake
```diff
@@ -0,0 +1,28 @@
...
+# The result is stored in a predefined output directory.
+#
+# USAGE 
+# 
+# The script can be used together with a custom_command or custom_target.
+# See example below: 
+# 
+# COMMAND ${CMAKE_COMMAND} -Dinput_directory=${input_directory} 
+#                          -Doutput_directory=${output_directory}
+#                          -Djunit_options=${options}
```
RReichert, line 15: by convention, all `-D` should be upper cased to imply global variable.

scripts/xml2junit_converter.py
```diff
@@ -0,0 +1,86 @@
...
+# framework supported by CI tools such as Jenkins to display test results.
+#
+# Run the script by supplying the file path to an xml file and an output
+# directory path as arguments.
+#
+import xml.etree.ElementTree as ET
```
RReichert, line 9: does this module come with python by default? I ask because of this https://github.com/swift-nav/swiftlets/pull/926#discussion_r479811611 other comment made in the past about python dependencies.

scripts/xml2junit_converter.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+#
+# This script converts an xml file into JUnit xml format. JUnit is a testing
+# framework supported by CI tools such as Jenkins to display test results.
+#
+# Run the script by supplying the file path to an xml file and an output
+# directory path as arguments.
```
RReichert, line 7: need a script usage description here to help future maintainers

JunitGenerator.cmake
```diff
@@ -0,0 +1,28 @@
...
+# be defined.
+#
+
+file(GLOB_RECURSE filenames ${input_directory}/*xml*)
+foreach(file_path ${filenames})
+  execute_process(
+    COMMAND python xml2junit_converter.py ${file_path} ${output_directory} ${junit_options}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/scripts
+  )
+endforeach()
```
RReichert, line 28: rather than doing this here and cmake script, why not move this logic into the python script? I say this because all the `*.cmake` files in the root folder can be `include()` into the project. Likewise I'm a bit concerned about the `file(GLOB_RECURSE)`, normally the results get cached and never updated on the next try and I don't know if that apply to running with the `cmake -P <script>` command.

scripts/xml2junit_converter.py
Davinco: Yes, It's part of the standard library: https://docs.python.org/3/library/xml.html

Valgrind.cmake
Davinco: Sure I can move it to `argsMult`. And an explanation is available here: https://github.com/swift-nav/cmake/blob/david/junit/Valgrind.cmake#L126-L129

Valgrind.cmake
Davinco: Hm I can't use `${CMAKE_CURRENT_SOURCE_DIR}`, that only gives me the path to where the test is located. However I found `${CMAKE_CURRENT_FUNCTION_LIST_DIR} that gave me the path to the cmake modules, but unfortunately that variable is only available from cmake version 3.17, so I'm going to keep `${CMAKE_SOURCE_DIR}/cmake/common` for now

Valgrind.cmake
RReichert: With the explanation that you've put up, I couldn't quite understand the `skip_test`, was that the actual keywork that you would need to add, or is that a place holder for something?

Valgrind.cmake
RReichert: Have you tried `CMAKE_CURRENT_LIST_DIR`? I forgot that `CMAKE_SOURCE_DIR` doesn't work when using `include()`, but I believe `CMAKE_CURRENT_LIST_DIR` should work.

Valgrind.cmake
```diff
@@ -311,13 +311,13 @@ function(swift_add_valgrind_memcheck target)
...
       COMMAND ${CMAKE_COMMAND} -E make_directory ${xml_dir}
-      COMMAND ${CMAKE_COMMAND} -Dinput_directory=${report_directory}/${report_folder} 
-                               -Doutput_directory=${xml_dir}
-                               -Djunit_options=${options} 
-                               -P ${CMAKE_SOURCE_DIR}/cmake/common/JunitGenerator.cmake
+      COMMAND ${CMAKE_COMMAND} -DScript=memcheck_xml2junit_converter.py
+                               -DInput_directory=${report_directory}/${report_folder}
+                               -DOutput_directory=${xml_dir}
+                               -DScript_options=${x_JUNIT_OPTIONS}
+                               -P ${CMAKE_SOURCE_DIR}/cmake/common/PythonWrapper.cmake
```
RReichert, line 320: what I was hoping for wasn't for a cmake wrapper, but passing `input_directory`, `output_directory`, and etc as arguments to the python script. have a look at https://github.com/swift-nav/swiftlets/blob/e68d16befd184aeb2849cf9df7afa7168de7f459/test/fast_start/run_fast_start_tests.py#L387, where they used such an approach. that way your script is independent of the build system and your script usage is well defined.

Valgrind.cmake
Davinco: Mm, `${CMAKE_CURRENT_LIST_DIR}` also only gives me the path to test's location

Valgrind.cmake
Davinco: I wanted to use the same structure as how other parameters are described, like `LEAK_CHECK` etc.
https://github.com/swift-nav/cmake/blob/david/junit/Valgrind.cmake#L106-L107
https://github.com/swift-nav/swiftlets/blob/david/memcheck-errors/test/unit/starling/CMakeLists.txt#L77

RReichert (approved):
LGTM :shipit:

PR #68, opened 2020-09-29
Add possibility to log total memory from profiling tools into a file

Davinco:
This is a step in the direction of comparing the difference in memory usage between two PR:s. My thought is to use https://docs.python.org/3/library/difflib.html on two log files and output the difference as a jenkins comment


Bloaty.cmake
```diff
@@ -179,4 +191,26 @@ function(swift_add_bloaty target)
...
+    set(memory_message "-m=Static memory usage:")
+
+    foreach (memory_option ${x_LOG_TOTAL_MEMORY_OPTIONS})
+      if (${memory_option} MATCHES "--input_file")
+        set(memory_input_dir ${memory_option})
+      elseif (${memory_option} MATCHES "--output_file")
+        set(memory_output_dir ${memory_option})
+      elseif (${memory_option} MATCHES "--message")
+        set(memory_message ${memory_option})
+      endif()
```
RReichert, line 207: There are a few things to comment on here:

1. `MATCHES` is a substring match, so if `memory_option` is for some reason set to `message=log--message` it would pick it up as a match, not sure if that is the full intention here
2. would be nice to have an else class here to alert of invalid potion
3. you could just forward the `message_option` directly to the script and have the script report errors or invalid option.

RReichert (commented):
Would you be able to create a PR that uses these changes, just as a check. Also I'm pretty bad at using python, so I cant give you critical comments on that

Davinco:
> Would you be able to create a PR that uses these changes, just as a check. Also I'm pretty bad at using python, so I cant give you critical comments on that

I've been using this PR as a test: https://github.com/swift-nav/swiftlets/pull/1068. However, the latest commit failed because of a permission issue so I need to check if something changed when using the groovy function `readFile`

Bloaty.cmake
Davinco: Good suggestions, I'll modify the scripts

RReichert (approved):

PR #71, opened 2020-11-10
[ESD-1894] Standardized compiler warning flags

RReichert:


benjaminaltieri (commented):
This looks like a solid approach. Will be great to have a sane default for general use. Can you link to prototype integration PRs in each of the major repos that would be consuming this? It helps to have examples of usage to understand the choice of cmake method design

RReichert:
> This looks like a solid approach. Will be great to have a sane default for general use. Can you link to prototype integration PRs in each of the major repos that would be consuming this? It helps to have examples of usage to understand the choice of cmake method design

There are a number of draft PRs using this:

* https://github.com/swift-nav/libixcom-private/pull/34
* https://github.com/swift-nav/libubx-private/pull/54
* https://github.com/swift-nav/librtcm-private/pull/115
* https://github.com/swift-nav/libswiftnav-private/pull/200

Unfortunately they are all still WIP, there have been a number of issues that have come trying to get this to work properly, depending on what compiler we use and what version, a number of false positives might come up, some that are compiler bugs all together (ex: mingw and its inability to handle `%zu` format), and just general developer debt to overcome. So these PRS I'll need to slowly work on over time.

PR #74, opened 2021-02-01
[ESD-1931] Common compiler flags

woodfell:
A long standing desire has been to have a common set of compiler flags for use with all Swift code. This has been tried several times in the past but never been finished, most recently by Rodrigo on whose work this PR is distantly based.

Rodrigo's work set compiler flags in the `LanguageStandards` module so that it would be automatically applied to everything that uses it. This is probably the way that it should be eventually, but to ease the transition I've broken it out to a separate module `CompileOptions` which can be applied to targets gradually until everything is up to date.

The set of compiler flags is taken from high level projects like starling and swiftlets. This means that low level libraries are now compiled with many more flags than they used to be (a good thing).

The exported function `swift_set_compile_options` can be called for any target. There are several options available for tailoring the set of options actually passed to the compiler, they should be fairly self explanatory.

All flags are verified against the compiler before actually being set. This is vital due to the massively diverse selection of compilers we have to cope with. Of  particular note is a CI stage in (I think) libswiftnav which uses clang for C sources and gcc for c++.

So far I've tested and verified this module with every project up to and including starling and PFWP. As part of this work I've also been updating the `.clang-tidy` config for each project (many low level libraries didn't have one at all). This means that once this module is merged and all submodules updates Swift code will be in such a state that we no longer need to set internal projects as system includes. There aren't any compiler warnings or clang-tidy errors.

This will remain open until everything else has been proven, which hopefully won't be too long.


CompileOptions.cmake
```diff
@@ -0,0 +1,183 @@
...
+      if (${cxx_enabled} GREATER -1)
+        check_cxx_compiler_flag(${flag} ${cxx_supported})
+        if(${${cxx_supported}})
+          list(APPEND final_flags $<$<COMPILE_LANGUAGE:CXX>:${flag}>)
+        endif()
+      endif()
+    endforeach()
+
+    foreach(target ${targets})
+      target_compile_options(${target} PRIVATE ${final_flags})
```
jbangelo, line 179: I don't mean to add more complexity here, but the compile flags like `-frtti` and `-fexceptions` should probably be marked as `PUBLIC` since it's not advisable to mix and match enabling and disabling those in a single executeable.

On the one hand we currently mark them as `PRIVATE`. This has caused strange issues in the past when making seemingly benign changes, and we've simply dealt with the symptoms and moved on without fixing the root problem. On the other hand allowing setting these specific flags to be `PUBLIC` will most likely make this much more complicated, and we've been making do with the wrong way so far.

I just wanted to point this out, in case we want to take on the added work (and risk), but I'm not sure that I would advocate for that (at least for right now).

jbangelo (approved):
Looks good to me. This will make things a lot more clean in our CMake files, and will help ensure we're more uniformly applying compiler options. Thanks for tackling this @woodfell !!

CompileOptions.cmake
woodfell: Yeah that's a good point. Added now along with some extra sanity checks and the ability to explicitly ask for no exceptions/rtti

woodfell: Actually, it looks like I'm going to have to take this out. Because of the way cmake constructs the compiler command line if a library which uses exceptions links against a library which doesn't `-fno-exceptions` will appear after `-fexceptions` and disable everything for the library which genuinely does need them. It won't be possible to just remove `-fno-exceptions` since exceptions are enabled by default. I ran in to this problem when updating libpal - the mocked pal-impl requires exceptions (because gmock does) but it links against libpal which explicitly disables exceptions. Unfortunate, but I think this is just one of those cmake things.

silverjam (approved):

PR #76, opened 2021-02-26
ESD-1959 allow for clang-tidy to be parallized by using the run-clang-tidy scr…

RReichert:
Made a change to the `clang-tidy` stage so that it prefers using the parallel script for launching clang-tidy static analysis over using the single threaded approach.

### DIFF
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -118,12 +118,12 @@ function(create_clang_tidy_targets)
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         )
     add_custom_target(clang-tidy-all-check
-        COMMAND test ! -f fixes.yaml
+        COMMAND test ! -s fixes.yaml
         DEPENDS clang-tidy-all
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         )
     add_custom_target(clang-tidy-diff-check
-        COMMAND test ! -f fixes.yaml
+        COMMAND test ! -s fixes.yaml
         DEPENDS clang-tidy-diff
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         )
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -248,6 +248,9 @@ function(swift_setup_clang_tidy)
   # First search for an appropriate clang-tidy
   if(NOT x_CLANG_TIDY_NAMES)
     set(x_CLANG_TIDY_NAMES
+        run-clang-tidy     run-clang-tidy.py
+        run-clang-tidy-6.0 run-clang-tidy-6.0.py
+        run-clang-tidy-4.0 run-clang-tidy-4.0.py
         clang-tidy60 clang-tidy-6.0
         clang-tidy40 clang-tidy-4.0
         clang-tidy39 clang-tidy-3.9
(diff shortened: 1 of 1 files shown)


ClangTidy.cmake
```diff
@@ -285,11 +288,11 @@ function(swift_setup_clang_tidy)
     create_clang_tidy_targets(
         TOP_LEVEL ${top_level_project}
         ALL_COMMAND
-        ${${PROJECT_NAME}_CLANG_TIDY} ${x_EXTRA_ARGS} -p ${CMAKE_BINARY_DIR} --export-fixes=${CMAKE_CURRENT_SOURCE_DIR}/fixes.yaml
+        ${${PROJECT_NAME}_CLANG_TIDY} ${x_EXTRA_ARGS} -p ${CMAKE_BINARY_DIR} -export-fixes=${CMAKE_CURRENT_SOURCE_DIR}/fixes.yaml
           `git ls-files ${srcs}`
         DIFF_COMMAND
         git diff --diff-filter=ACMRTUXB --quiet --name-only master -- ${srcs} ||
-        ${${PROJECT_NAME}_CLANG_TIDY} ${x_EXTRA_ARGS} -p ${CMAKE_BINARY_DIR} --export-fixes=${CMAKE_CURRENT_SOURCE_DIR}/fixes.yaml
+        ${${PROJECT_NAME}_CLANG_TIDY} ${x_EXTRA_ARGS} -p ${CMAKE_BINARY_DIR} -export-fixes=${CMAKE_CURRENT_SOURCE_DIR}/fixes.yaml
```
logingood, line 295: what about `--export-fixes` this example says double `--` https://clang.llvm.org/extra/clang-tidy/

RReichert: This issue caught me off guards. If you spin up the `swift-build` docker image, and you run `clang-tidy-6.0 --help`, you surprisingly get:

```
USAGE: clang-tidy-6.0 [options] <source0> [... <sourceN>]

OPTIONS:

Generic Options:

  -help                        - Display available options (-help-hidden for more)
  -help-list                   - Display list of available options (-help-list-hidden for more)
  -version                     - Display the version of this program

clang-tidy options:
...
  -export-fixes=<filename>     - 
                                 YAML file to store suggested fixes in. The
                                 stored fixes can be applied to the input source
                                 code with clang-apply-replacements.
...
```
And if I check my laptop's clang-tidy help text, I get what you see in the website you've provided. Then when you look at `run-clang-tidy.py --help` script, which is almost a stand in replacement for clang-tidy it provides:
```
usage: run-clang-tidy.py [-h] [-clang-tidy-binary PATH]
                         [-clang-apply-replacements-binary PATH]
                         [-checks CHECKS] [-config CONFIG]
                         [-header-filter HEADER_FILTER]
                         [-export-fixes filename] [-j J] [-fix] [-format]
                         [-style STYLE] [-p BUILD_PATH] [-extra-arg EXTRA_ARG]
                         [-extra-arg-before EXTRA_ARG_BEFORE] [-quiet]
                         [files [files ...]]

Runs clang-tidy over all files in a compilation database. Requires clang-tidy
and clang-apply-replacements in $PATH.

positional arguments:
  files                 files to be processed (regex on path)

optional arguments:
...
  -export-fixes filename
                        Create a yaml file to store suggested fixes in, which
                        can be applied with clang-apply-replacements.
...
```
But surprisingly, `clang-tidy` on my machine still works with the older `-export-fixes` option, even if they don't advertise it, I guess its for backwards compatibility.

ClangTidy.cmake
```diff
@@ -118,12 +118,12 @@ function(create_clang_tidy_targets)
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         )
     add_custom_target(clang-tidy-all-check
-        COMMAND test ! -f fixes.yaml
+        COMMAND test ! -s fixes.yaml
```
RReichert, line 121: with `run-clang-tidy` script, it always produces a `fixes.yaml`, irrespective of whether or not there is an error, so now it tests to make sure to fail if there a file AND its size is greater than zero.

ClangTidy.cmake
logingood: interesting, weird though

logingood (approved):

ClangTidy.cmake
woodfell: How long until someone checks in an empty `fixes.yaml`...? :smile:

ClangTidy.cmake
```diff
@@ -248,6 +248,9 @@ function(swift_setup_clang_tidy)
   # First search for an appropriate clang-tidy
   if(NOT x_CLANG_TIDY_NAMES)
     set(x_CLANG_TIDY_NAMES
+        run-clang-tidy     run-clang-tidy.py
+        run-clang-tidy-6.0 run-clang-tidy-6.0.py
+        run-clang-tidy-4.0 run-clang-tidy-4.0.py
```
woodfell, line 253: We can probably just ditch any version other than 6. This is only here fot legacy reasons and lower versions probably won't work properly.

woodfell (approved):
Nice

RReichert:
This PR needed to be removed from master for a couple of reasons:

1. was failing with libpal because unlike `clang-tidy`, the files passed onto `run-clang-tidy` script are interpreted as regular expressions, so it fails with `pal++` folder
2. in starling the reason its taking longer than expected is because some files are being run twice, don't know

Anyways the attempt of trying to avoid creating a new function and just piggy back off of the old PR is not working, will need to rewrite the implementation as a new PR

PR #79, opened 2021-03-31
ESD-1981 code coverage overhaul

RReichert:
### Problem

Starling's CI is a bit of a mess at the moment, there are some stages that run some unit tests, others that don't run it at. This PR is hoping to ease some of these issues by breaking down the tests into two categories and having the CI stages simply call `do-all-unit-tests` or `do-all-integration-tests` (the latter might not be possible at the moment because some of the integration tests take too long and running them all will take ages, but the former certainly is). This isn't only useful for CI, its required by ISO 26262, it asks that we provide code coverage reports for both test types, doing this makes things a lot easier.

### PR Changes

This PR shouldn't break anything, it will only provide the means for tests to be flagged as either "Unit Test" or "Integration Test". Any existing function calls will now get a cmake warning asking users to specify it, but it won't break anything.

I also added the following new feature to the PR:
```
If the PARALLEL option was specified for swift_add_test, than the unit tests
will be run in parallel when executing do-all-tests.
```
This is done to speed up our builds

### Related Branch

I've got a branch that needs this PR updated across the organization before it works. I've got it working on my computer by updating the common cmake submodule.

https://github.com/swift-nav/starling/compare/rodrigor/ESD-1981-code-coverage-overhaul


TestTargets.cmake
```diff
@@ -96,6 +113,8 @@
 #    COMMAND <path to executable>
 #    )
 #
+# NOTE: using POST_BUILD option is not advised as it will increase build time
```
RReichert, line 116: added this as I'm working on removing post build testing in starling as its doubling up on testing and wasting $$ every build.

TestTargets.cmake
```diff
@@ -46,11 +52,13 @@
 # from the googletest package. The non-parallel target is always created.
 #
 # The other function, swift_add_test_runner(), can be used to create a do-... 
-# target which pointer at some other command. This can be used to invoke a test 
+# target which points at some other command. This can be used to invoke a test
 # which is actually a shell script or some other executable which is available 
 # but not built from source
 #
 # swift_add_test_runner(test-suite-name
+#    UNIT_TEST
```
woodfell, line 60: Perhaps indicate that these are mutually exclusive

woodfell (approved):

PR #82, opened 2021-06-05
[ESD-2055] Disable extensions in C code, allow deprecated declarations at warning leve

woodfell:
For some reason the language standards module enable C language extensions by default. This is causing gcc to be passed `-std=gnu11` rather than `-std=c11`

Alter the set of compiler warning to allow the use of deprecated functions without causing a compile error regardless of other options in use.


RReichert (approved):
LGTM

samvrlewis (approved):

PR #93, opened 2021-07-19
[ESD-2038] Script to check for misused of __attribute__

woodfell:
The use of plain `__attribute__` is causing some portability issues. https://github.com/swift-nav/libswiftnav-private/pull/274 introduces a header to abstract away attributes behind some preprocessor macros which are protected and redefined by compiler detecting `#if` blocks

This script is a parallel effort to help prevent future erroneous usage of `__attribute__`. This module and script can be used by any project and adds a new target which looks through all source and header files in a repository and flags an IDE parsable error should `__attribute__` be used anywhere. The script is fairly dumb and just greps through all files tracked by git. It's not a perfect solution but will definitely fail CI in the event of a misuse.

This has been demonstrated in other Swift repositories in a series of other PRs

### DIFF
--- a/CheckAttributes.cmake
+++ b/CheckAttributes.cmake
@@ -1,0 +1,8 @@
+if(TARGET check-attributes)
+  return()
+endif()
+
+add_custom_target(check-attributes ALL
+  ${CMAKE_CURRENT_SOURCE_DIR}/cmake/common/scripts/check_attributes.sh
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  )
--- a/CheckAttributes.cmake
+++ b/CheckAttributes.cmake
@@ -22,7 +22,7 @@ function(create_check_attributes_target)
 
   cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
 
-  set(arguments "'*.[ch]'" "'*.[ch]pp'" "'*.cc'")
+  set(arguments "'*.[ch]'" "'*.[ch]pp'" "'*.cc'" "'*.[ch]xx'")
   foreach(excl ${x_EXCLUDE})
     list(APPEND arguments ":!:${excl}")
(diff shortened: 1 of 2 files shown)


CheckAttributes.cmake
```diff
@@ -0,0 +1,35 @@
...
+    return()
+  endif()
+
+  set(argOption "")
+  set(argSingle "")
+  set(argMulti "EXCLUDE")
+
+  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
+
+  set(arguments "'*.[ch]'" "'*.[ch]pp'" "'*.cc'")
```
silverjam, line 25: Should we ".cxx" and ".hxx" too?

scripts/check_attributes.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+#
+
+failed=0
```
silverjam, line 14: Suggest just naming this `status`

silverjam (changes requested):

silverjam (commented):
Also suggest running this script through shellcheck: https://github.com/koalaman/shellcheck

woodfell:
Reworked the script to get it passing with `shellcheck` while still not requiring bash

RReichert (approved):
LGTM, if you can be bothered, a quick doc on the function  would be nice to have

PR #94, opened 2021-07-22
ESD-2131 swift targets

RReichert:
## Changes

I've created a new `SwiftTargets.cmake` file which users can include in their root `CMakeLists.txt` in order to expose the following new functions:

* `swift_add_executable` -  will replace `add_executable`, executables that are production code should be created through this function
* `swift_add_library` - will replace `add_library`, libraries that are production code should be created through this function
* `swift_add_test_library` -  will replace `add_library`, libraries that aren't production code but are used for testing should be create through this function.

These functions help to achieve the following goals:

* mark which targets are ours and categorize them
* prevent issues where an executable/library doesn't also call `swift_set_language_standards` or `swift_set_compile_options` or `target_code_coverage` which are needed across the organization for various reasons
* centralize the control of our targets

We currently have an issue in a number of our repositories where there are targets which aren't conforming to the coding standards because they aren't calling the `swift_set_compile_options` or they don't have code coverage data because they haven't invoked `target_code_coverage`.

## Design Considerations

I'm aware the we also have the `swift_add_test` function which has a different API to `swift_add_test_library`. As as a user I would expect there to be some common API between the two, which I've decided not to do. The decision again this is mostly due to the fact that the `LINK` and `INCLUDE` options will over complicate the implementation, this will lead to an API much [`install`](https://cmake.org/cmake/help/latest/command/install.html) that has nested options. Unlike `swift_add_test`, `swift_add_test_library` (along with `swift_add_executable` and `swift_add_library`) can link again libraries in a PRIVATE/PUBLIC/INTERFACE manner, this is currently not possible with `swift_add_test` as it assumes they are all `PRIVATE` (for good reasons). Given that we have no requirements right now to do any special with the `INCLUDE` or `LINK`, I'd like to avoid introducing it to the interface right now, I'd like for us to consider either removing the options from `swift_add_test`'s in the future or upgrade these new functions to include them with the additional option to specify PRIVATE/PUBLIC/INTERFACE.


SwiftTargets.cmake
```diff
@@ -0,0 +1,113 @@
...
+  swift_add_target("${target}" tool_library ${ARGN})
+endfunction()
+
+function(swift_add_library target)
+  swift_add_target("${target}" library ${ARGN})
+endfunction()
+
+function(swift_add_test_library target)
+  swift_add_target("${target}" test_library ${ARGN})
+endfunction()
```
jayvdb, line 113: note CI will now require this to have an EOL at EOF

RReichert: made the change

SwiftTargets.cmake
```diff
@@ -0,0 +1,283 @@
...
+  if (x_SHARED)
+    list(APPEND library_type SHARED)
+  endif()
+  if (x_MODULE)
+    list(APPEND library_type MODULE)
+  endif()
+
+  if (type STREQUAL "executable")
+    add_executable(${target} ${x_SOURCES})
+  elseif(type STREQUAL "library")
```
woodfell, line 234: As mentioned in the LSNP PR perhaps add an alias somewhere in here

woodfell (approved):
Looks great. Can you please leave a ticket in the tech debt epic to come back to everything mentioned in the PR description to do with the test targets API

RReichert:
Tech Debt: https://swift-nav.atlassian.net/browse/ESD-2135

SwiftTargets.cmake
RReichert: added in tech debt ticket

SwiftTargets.cmake
```diff
@@ -0,0 +1,283 @@
...
+#
+# For those that just want to get a quick synopsis of what it takes to use the
+# function, lets say we want to create a new library, lets call it "my-lib",
+# you can do so via:
+#
+#   swift_add_library(my-lib SOURCES a.cc b.cc c.cc)
+#
+# This would be roughly equivalent to:
+#
+#   add_librar(my-lib a.cc b.cc c.cc)
```
jayvdb, line 71: sp `add_library`

SwiftTargets.cmake
```diff
@@ -0,0 +1,283 @@
...
+#   * swift_add_tool
+#
+# There are a number of keywords you can use for the functions, they are listed
+# below. Just a reminder that when using these keywords, please try and specify
+# the multi-value keywords last, this can avoid some of the cmake issues that
+# people might stumble upon if they incorrectly spell the keyword.
+#
+# OPTION KEYWORDS
+#
+#  INTERFACE: this only works for library targets, it marks the library as an
```
jayvdb, line 96: `s/this/This/`? & `s/,/;/` (or .) otherwise this (and the next paragraph) are run-on non-sentences?

SwiftTargets.cmake
```diff
@@ -0,0 +1,283 @@
...
+# OVERVIEW
+# ========
+#
+# Offers a set of functions which should be replacements for "add_executable"
+# and "add_library". Upon including this module, the following functions should
+# be available to users:
+#
+#   * swift_add_executable - defines an executable which is production worthy
+#   * swift_add_library - defines a library which is production worthy
+#   * swift_add_tool - defines an executable which is an internal tool
```
jayvdb, line 23: would this be used for example executables? e.g. https://github.com/swift-nav/libsbp/tree/master/c/examples

SwiftTargets.cmake
```diff
@@ -0,0 +1,283 @@
...
+# set forth by functions like "swift_set_compile_options" and
+# "swift_set_language_standards".
+#
+# Note that by including this module, the following functions are also
+# available for use:
+#
+#   * swift_add_test
+#   * swift_add_test_runner
+#
+# These functions are defined in "SwiftTargets.cmake", so please consult that
```
jayvdb, line 37: `TestTargets.cmake` ?

SwiftTargets.cmake
RReichert: Probably not, examples aren't that important to track and enforce coding compliance which is why this wasn't included, I've added to the tech debt ticket.

SwiftTargets.cmake
RReichert: good spotting, took care of that

PR #96, opened 2021-07-26
CI: Use cmake-lint

jayvdb:
https://cmake-format.readthedocs.io/en/latest/index.html also has a formatter that we could use, however it will need a bit more tuning, and it requires more invasive changes to the .cmake files.


silverjam (approved):
Looks fine but would definitely wait for @woodfell or @RReichert to approve

.github/workflows/lint.yaml
```diff
@@ -9,14 +9,18 @@ on:
...
+        run: pip install trim cmakelang
 
       - name: Check whitespace
         run: |
           trim *.cmake
           git diff --exit-code
+
+      - name: Lint cmake files
+        run: |
+          cmake-lint *.cmake
```
RReichert, line 26: do we need to add `git diff --exit-code` here to fail on incorrectly formatted cmake files?

.cmake-format.yaml
```diff
@@ -0,0 +1,63 @@
+parse: {}
+format:
+  line_width: 227
```
RReichert, line 3: nit and open question, should this be relative to some of our other formatting files, [`clang-format` defines 80 for column](https://github.com/swift-nav/starling/blob/4798cff7d1bedd708c0b8a25c081a206187ec778/.clang-format#L51)

RReichert (commented):
I think this is in the right direction, the only concern I have (like MOST of our formatting yaml files) is trying to centralize the format file. This is only if we want to expand this out to other repos (say we want to format the various `CMakeLists.txt` file). I'm happy with the current settings and if we stumble upon any issues, we can update it later on.

The only thing currently blocking me from approving is the question I have about the `git diff` below. Once that is resolved I'm happy to give the thumbs up.

Alt marie kondo meme

.github/workflows/lint.yaml
jayvdb: No.  cmake-lint doesnt reformat.  There is a separate cmake-format which does, and then we'd need `git diff --exit-code`.  It will involve quite a bit more work, and less gain as the overall format is ok.

.cmake-format.yaml
jayvdb: cmake-lint does default to 80.  IMO we might want to use 120 here.  But we have files that have 227 chars.  I could easily bring this down to 180-ish by splitting the really long lines, or by adding cmake-lint pragmas on specific files.  Probably could get to a default of 120 without too much effort.  Trying to reach 80 would be painful.

.github/workflows/lint.yaml
RReichert: Would be worth having a small entry in the `README.txt` to explain how to format the files

.cmake-format.yaml
RReichert: I'm happy for 120 or 180 (I also feel 80 is quite small). Happy to use whatever seems nice, just happy to have some linting at this point. Would be worth seeing the effect of what formatting does if it flows beyond the threshold.

RReichert (approved):

CodeCoverage.cmake
```diff
@@ -82,7 +82,9 @@
 # add_code_coverage_all_targets(EXCLUDE test/*) # Adds the 'ccov-all' target set and sets it to exclude all files in test/ folders.
 #
 # add_executable(theExe main.cpp non_covered.cpp)
-# target_code_coverage(theExe AUTO ALL EXCLUDE non_covered.cpp test/*) # As an executable target, adds to the 'ccov' and ccov-all' targets, and the reports will exclude the non-covered.cpp file, and any files in a test/ folder.
+# target_code_coverage(theExe AUTO ALL EXCLUDE non_covered.cpp test/*)
+#          # As an executable target, the above adds to the 'ccov' and ccov-all' targets,
```
RReichert, line 86: nit, comments are badly formatted due to wrapping

.cmake-format.yaml
jayvdb: Changed to 180, and I'll let more experienced cmake devs drive this down to 120.  `# cmake-lint: disable=C0301` disables line length for the entire file, so best not to be used atm , c.f. https://github.com/cheshirekow/cmake_format/issues/271

CodeCoverage.cmake
jayvdb: happy to amend it; how would you prefer it ?

RReichert: my fault, I thought that the line was a single long sentence that got wrapped incorrectly, but I see that the purpose of the line was to have a comment at the end of the `target_code_coverage`. just ignore my comment, what you've done will do.

.cmake-format.yaml
RReichert: Happy with the change

PR #101, opened 2021-09-24
[ESD-2180] Improved clang-tidy module

woodfell:
There is a bit going on in here.

First there is a new module `ListTargets.cmake` which can enumerate all targets in a source tree based on certain criteria

Second there is a new property defined on targets `SWIFT_TYPE` which is filled in by all the `swift_add_*` functions. This property is one of the criteria which ListTargets can filter on.

There is a new function `swift_validate_targets` which combines the above two items to make sure that all compilable targets defined in a project were added with `swift_add_*` rather than the native cmake `add_*`. In the future this function can be expanded to provide extra validation of targets.

Finally there is a completely rewritten `ClangTidy` module. Rather than having to pass a git GLOB pattern to `swift_setup_clang_tidy` a replacement function `swift_create_clang_tidy_targets` uses ListTargets and the new property to be far more smart about building up the list of source files to pass to clang-tidy

All targets defined within a repository have their own clang tidy targets, eg `clang-tidy-starling`, `clang-tidy-pvt-engine`. Fixes are exported to `fixes-<target>.yaml`

`clang-tidy-all` still exists and lints all "core" parts of starling, anything defined with `swift_add_executable` or `swift_add_library`

`clang-tidy-world` lints absolutely everything defined in a repository, all core targets, tools, tests, test libraries and so on.

Each `clang-tidy-*` target has a corrosponding `clang-tidy-<target>-check` target which returns an error code should any errors be generated.

~The old "diff" clang-tidy targets no longer exist.~

All clang-tidy targets are run in parallel using a script from the LLVM project.

`.clang-tidy` is now autogenerated during `swift_setup_clang_tidy`, as this module is updated in other repositories all versions of `.clang-tidy` can be removed and we can finally have proper standardisation across key projects. In recognition that some other projects really do want their own configs this behaviour is controllable and can be disabled if required.


scripts/run-clang-tidy.py
```diff
@@ -0,0 +1,344 @@
+#!/usr/bin/env python
```
woodfell, line 1: This file taken almost verbatim from LLVM.
The only change is to not write an empty fixes file if no errors were generated

ListTargets.cmake
```diff
@@ -0,0 +1,116 @@
...
+#
+# A helper module to enumerate all targets created in a source tree
+#
+# Calling swift_list_targets will populate the given variable with the list of targets which meet the given criteria. It will report all targets created in the current source dir
+# and all of its subdirectories
+#
+# Use the ONLY_THIS_REPO option to exclude targets which were created in submodules, returns a list of targets which were created in the caller's repository. This is fairly dumb
+# test which simply looks for 'third_party' anywhere in the path of the candidate target and excludes it
+#
+# Use the TYPES option to filter the returned list of targets based on the cmake target type. Only targets which match one of the entries in this list will be returned
```
RReichert, line 19: nit: maybe add a link to what options are possible here

ListTargets.cmake
```diff
@@ -0,0 +1,116 @@
...
+# Populate 'interfaces' with the set of interface library targets created within this repository
+#
+# swift_list_targets(interfaces ONLY_THIS_REPO TYPES "INTERFACE")
+#
+# Populate 'all_tests' with the set of tests defined anywhere within the source tree under the calling function's directory
+#
+# swift_list_targets(all_tests SWIFT_TYPE "test")
+#
+# A helper function swift_list_compilable_targets is provided which is equivalent to calling swift_list_targets with the option "TYPES EXECUTABLE DYNAMIC_LIBRARY STATIC_LIBRARY
+# OBJECT_LIBRARY", ie the set of all targets which can be compiled (not interface libraries). Other options can be passeed in to this function (ONLY_THIS_REPO, SWIFT_TYPES) as for
```
RReichert, line 35: nit: `passeed` -> `passed`

ClangTidy.cmake
```diff
@@ -1,308 +1,216 @@
...
-#
-# The default value is ON for top level projects, and OFF for any others.
-#
-# Running
-#
-# cmake -D<project>_ENABLE_CLANG_TIDY=OFF ..
-#
-# will explicitly disable these targets from the command line at configure time
+# swift_create_clang_tidy_targets will generate a .clang-tidy file in the project source directory which contains the Swift master config for clang-tidy. There is no need for
+# repositories to maintain their own version of .clang-tidy, it should be added to .gitignore in each repository to prevent being checked in. Pass the option
```
RReichert, line 36: nice, like the idea. I'd probably ask on slack if anyone relies on the `.clang-tidy` on their repo. I personally dont, but I know that CLion does have some nice integration with clang-tidy and reports these errors in realtime in the IDE.

ClangTidy.cmake
```diff
@@ -1,308 +1,216 @@
...
+  set(CMAKE_EXPORT_COMPILE_COMMANDS
+      ON
+      CACHE BOOL "Export compile commands" FORCE)
 
-  cmake_parse_arguments(x "${argOption}" "${argSingle}" "${argMulti}" ${ARGN})
-
-  if(x_UNPARSED_ARGUMENTS)
-    message(FATAL_ERROR "Unparsed arguments ${x_UNPARSED_ARGUMENTS}")
-  endif()
+  find_program(CLANG_TIDY NAMES clang-tidy-6.0 clang-tidy)
```
RReichert, line 73: :tada: clang-tidy-4.0 is finally gone

ClangTidy.cmake
```diff
@@ -1,308 +1,216 @@
...
+        -readability-redundant-declaration
+        -readability-redundant-member-init)
+
+    # Final list of checks to enable/disable
+    set(all_checks -* ${enabled_categories} ${disabled_checks})
+
+    # .clang-tidy needs a comma separated list of checks but cmake uses a semicolon as the list separator
+    string(REPLACE ";" "," comma_checks "${all_checks}")
+
+    # Generate .clang-tidy in project root dir
```
RReichert, line 165: cmake has a not very well known "heredoc" like syntax if you are [interested](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#bracket-argument) and find that it might help with readability

RReichert (approved):
Really awesome work. Minor nits and comments, but looks good to me.

isaactorz (approved):
LGTM.

ClangTidy.cmake
```diff
@@ -1,308 +1,216 @@
 #
-# Copyright (C) 2021 Swift Navigation Inc.
-# Contact: Swift Navigation <dev@swift-nav.com>
+# Copyright (C) 2021 Swift Navigation Inc. Contact: Swift Navigation <dev@swift-nav.com>
```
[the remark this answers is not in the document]
woodfell: No, I made the mistake of running `cmake-format` before committing everything. It cocked up the entire file. I'll revert this header block but won't bother with the rest of the file even though it looks like garbage

ClangTidy.cmake
woodfell: Yeah I use something similar. It's pretty important to keep `.clang-tidy` in the repo root thanks to the braindead way `clang-tidy` looks for its config. It would be so much easier if you could pass the path to a config file as a command line argument, but for whatever reason they didn't implement it like that.

ClangTidy.cmake
woodfell: Looks like it doesn't do variable expansion which is required here, not sure it would really change much

woodfell: ... although it seems like a simple double quoted multi line string would do the job

woodfell:
I've had to add the original ClangTidy.cmake back in to this PR (as OldClangTidy.cmake) due to a problem in some higher level repositories that haven't been converted over to use the `swift_add_*` functions yet. It was going to be too much effort to convert all of them as well (and probably cause a fair amount of annoyance due to stricter compile options) so they can keep the old module for the moment.

PR #104, opened 2021-11-02
[ESD-2251] Automatically set compile options when creating tests

woodfell:
I'm constantly forgetting to call `swift_set_compile_options` after creating a test which very often leads to linker errors due to mismatched RTTI flags.

Always call `swift_set_compile_options` with no arguments when creating a test. This doesn't prevent this function from being called separately should the creator of a test really need to customise compiler flags.

This is pretty much the same thing as already happens for `swift_set_language_standards`


isaactorz (commented):
Does it matter if the target is a "tool" ?

isaactorz (approved):

silverjam:
@woodfell merge?

PR #105, opened 2021-11-10
[ESD-2257] Move archive util to common cmake repo

woodfell:
This was previously implemented in a different repo and ended up being used from a different one. It really belongs here


isaactorz (approved):
What a pain that cmake doesn't provide a way to handle this out the box.

Good solution though.

woodfell:
Yep there's lot of little things that would be really nice if cmake had a built in way of doing

PR #107, opened 2021-12-02
fix: use correct target types for compilable targets

abdes:
Function ``swift_list_compilable_targets`` in `ListTargets,cmake` filters targets using target types that are not the ones returned by cmake ``get_target_property(TYPE)``.

The correct list should be:
```
    "EXECUTABLE"
    "MODULE_LIBRARY"
    "SHARED_LIBRARY"
    "STATIC_LIBRARY"
    "OBJECT_LIBRARY"
```


woodfell (approved):

isaactorz (approved):
LGTM

PR #111, opened 2022-05-06
Integrity unit tests framework

RReichert:
# Background

In looking to introduce unit/integration unit testing into Orion Engine for the Bosch integrity work, I've stumbled up on a few issues in Orion and friends. Firstly there aren't any `find_packge` scripts for `orion-engine`, `orion_proto`, `sip`, and `albatros` in this cmake repository, that isn't a problem since each repository has a custom script to include them in the `cmake/` directory rather than the `cmake/common`. The problem is that each of these scripts have not included the necessary `option(${PROJECT_NAME}_ENABLE_${FEATURE} "" false)` statements which is needed to control the inclusion of submodule test executable. 

This causes issues when running `do-all-tests` in `orion-engine` as unit tests from both `sip` (uses [swift_add_tests](https://github.com/swift-nav/sip/blob/0bed20094eda02a2b14aebbfbf9150144fd415e0/CMakeLists.txt#L256)) and `starling` (ex: the inclusion of [test-starling-output-analyzer](https://github.com/swift-nav/starling/blob/8bc2ba6309ce5eee08ed55cdf5d00d2d18b6e8a9/tools/starling-output-analyzer/test/unit/CMakeLists.txt#L13) isn't controlled by `starling_BUILD_TESTS` [here](https://github.com/swift-nav/starling/blob/8bc2ba6309ce5eee08ed55cdf5d00d2d18b6e8a9/tools/starling-output-analyzer/CMakeLists.txt#L49)) are executed. So to fix this, this PR does two things:

1. introduce the find_package scripts for the missing Swift repositories and marks all doc/tests/example as false
2. fixes up the `swift_add_test` and `swift_add_test_runner` functions so that even if the targets aren't protected by a `${PROJECT}_BUILD_TESTS` variable (and variants), it won't be included into the `do-all-tests` target.

# Testing

Ran the following script in Orion and than tried to compile the submodules (orion-engine, sip, orion_proto) after I removed the local find_package scripts.  

```shell
#!/usr/bin/bash

REPOSITORIES=(
    cmake/common/
    third_party/orion_proto/cmake/common/
    third_party/orion_proto/third_party/gnss-converters/c/cmake/common/
    third_party/orion_proto/third_party/gnss-converters/c/third_party/libsbp/c/cmake/common/
    third_party/orion_proto/third_party/gnss-converters/c/third_party/libswiftnav/cmake/common/
    third_party/orion_proto/third_party/sip/cmake/common/
    third_party/orion_proto/third_party/sip/third_party/libsbp/c/cmake/common/
    third_party/orion_proto/third_party/sip/third_party/libswiftnav/cmake/common/
    third_party/orion-engine/cmake/common/
    third_party/orion-engine/third_party/orion_proto/cmake/common
    third_party/orion-engine/third_party/orion_proto/third_party/gnss-converters/c/cmake/common/
    third_party/orion-engine/third_party/orion_proto/third_party/gnss-converters/c/third_party/libsbp/c/cmake/common/
    third_party/orion-engine/third_party/orion_proto/third_party/gnss-converters/c/third_party/libswiftnav/cmake/common/
    third_party/orion-engine/third_party/orion_proto/third_party/sip/cmake/common/
    third_party/orion-engine/third_party/orion_proto/third_party/sip/third_party/libsbp/c/cmake/common/
    third_party/orion-engine/third_party/orion_proto/third_party/sip/third_party/libswiftnav/cmake/common/
    third_party/orion-engine/third_party/starling/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/libpal/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/libsbp/c/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/libswiftnav/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/gnss-converters/c/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/gnss-converters/c/third_party/libsbp/c/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/gnss-converters/c/third_party/libswiftnav/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/libpal/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/libsbp/c/cmake/common/
    third_party/orion-engine/third_party/starling/third_party/swiftlets/third_party/libswiftnav/cmake/common/
)

for repository in ${REPOSITORIES[@]}
do
  git -C ${repository} fetch
done


for repository in ${REPOSITORIES[@]}
do
  git -C ${repository} checkout origin/rodrigor/integrity-unit-tests-framework
done
```

Also to verify that this won't have any impact on Starling, I created [this PR](https://github.com/swift-nav/starling/pull/6545).

### DIFF
--- a/FindAlbatross.cmake
+++ b/FindAlbatross.cmake
@@ -12,12 +12,12 @@
 
 include("GenericFindDependency")
 
-option(orion_proto_ENABLE_DOCS "" false)
-option(orion_proto_ENABLE_EXAMPLES "" false)
-option(orion_proto_ENABLE_TESTS "" false)
-option(orion_proto_ENABLE_TEST_LIBS "" false)
+option(albatross_ENABLE_DOCS "" false)
+option(albatross_ENABLE_EXAMPLES "" false)
+option(albatross_ENABLE_TESTS "" false)
+option(albatross_ENABLE_TEST_LIBS "" false)
 
 GenericFindDependency(
-  TARGET orion-proto
-  SOURCE_DIR orion_proto
+  TARGET albatross
+  SYSTEM_INCLUDES
 )
--- a/FindOrion-Engine.cmake
+++ b/FindOrion-Engine.cmake
@@ -12,12 +12,13 @@
 
 include("GenericFindDependency")
 
-option(orion_proto_ENABLE_DOCS "" false)
-option(orion_proto_ENABLE_EXAMPLES "" false)
-option(orion_proto_ENABLE_TESTS "" false)
-option(orion_proto_ENABLE_TEST_LIBS "" false)
+option(orion-engine_ENABLE_DOCS "" false)
+option(orion-engine_ENABLE_EXAMPLES "" false)
+option(orion-engine_ENABLE_TESTS "" false)
+option(orion-engine_ENABLE_TEST_LIBS "" false)
 
 GenericFindDependency(
-  TARGET orion-proto
-  SOURCE_DIR orion_proto
+  TARGET orion_engine
+  SOURCE_DIR orion-engine
+  SYSTEM_INCLUDES
 )
--- a/FindSip.cmake
+++ b/FindSip.cmake
@@ -12,12 +12,13 @@
 
 include("GenericFindDependency")
 
-option(orion_proto_ENABLE_DOCS "" false)
-option(orion_proto_ENABLE_EXAMPLES "" false)
-option(orion_proto_ENABLE_TESTS "" false)
-option(orion_proto_ENABLE_TEST_LIBS "" false)
+option(sip_ENABLE_DOCS "" false)
+option(sip_ENABLE_EXAMPLES "" false)
+option(sip_ENABLE_TESTS "" false)
+option(sip_ENABLE_TEST_LIBS "" false)
 
 GenericFindDependency(
-  TARGET orion-proto
-  SOURCE_DIR orion_proto
+  TARGET sip
+  SOURCE_DIR sip
+  SYSTEM_INCLUDES
 )
--- a/TestTargets.cmake
+++ b/TestTargets.cmake
@@ -197,8 +197,6 @@ function(swift_add_test_runner target)
     message(FATAL_ERROR "swift_add_test_runner unparsed arguments - ${x_UNPARSED_ARGUMENTS}")
   endif()
 
-  swift_create_test_targets()
-
   if(NOT x_COMMENT)
     set(x_COMMENT "test ${target}")
   endif()
@@ -212,6 +210,12 @@ function(swift_add_test_runner target)
   elseif(x_INTEGRATION_TEST AND x_UNIT_TEST)
     message(FATAL_ERROR "Both INTEGRATION_TEST and UNIT_TEST option were specified, you can only specify one")
   endif()
+
+  if (NOT ${PROJECT_NAME}_BUILD_TESTS)
+    return()
+  endif()
+
+  swift_create_test_targets()
 
   add_custom_target(
     do-${target}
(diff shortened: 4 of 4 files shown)


FindAlbatross.cmake
```diff
@@ -0,0 +1,23 @@
...
+include("GenericFindDependency")
+
+option(albatross_ENABLE_DOCS "" false)
+option(albatross_ENABLE_EXAMPLES "" false)
+option(albatross_ENABLE_TESTS "" false)
+option(albatross_ENABLE_TEST_LIBS "" false)
+
+GenericFindDependency(
+  TARGET albatross
+  SYSTEM_INCLUDES
```
RReichert, line 22: these have been included in the customer scripts, I'm not in favor of doing this since this is our code and we should be linting, but I'm not going to attempt to open that can of worms right now

isaactorz: A little unclear what you mean here. We shouldn't be setting the options in our find scripts?

isaactorz (approved):
lgtm but one question.

What's the purpose for having both an ENABLE_TESTS and a BUILD_TESTS option?

FindAlbatross.cmake
RReichert: @isaactorz We should only be using `SYSTEM_INCLUDE` for code that doesn't comply with our coding standards. For example, the optional library ([see here](https://github.com/swift-nav/cmake/blob/a4f91f7bc645d65b30177f057be5db7e91310c44/FindOptional.cmake#L17)) is something that we've imported into our code base from an external source and it might not comply with our strict coding guidelines. Swiftlets on the other hand is something that we internally developed ([see here](https://github.com/swift-nav/cmake/blob/a4f91f7bc645d65b30177f057be5db7e91310c44/FindSwiftlets.cmake#L18)) and therefore must comply.

Internally this `SYSTEM_INCLUDES` flag makes sure that all header files public from these third party targets are included with `-isystem` when compiling with GCC and Clang rather than the standard `-I`. So for example, if we wanted to link `optional`, it would compile it with something like this:

```shell
gcc -o main main.cc ... -L third_party/optional -loptional -isystem /third_party/optional/include ...
```
rather than the usual

```shell
gcc -p main main.cc ... -L third_party/optional -loptional -I /third_party/optional/include ...
```

Any warnings from header files under  `-isystem` would be ignored rather than cause an error during compilation. Likewise the `clang-tidy` would ignore any issues under `-isystem`.

RReichert:
@isaactorz 

> lgtm but one question.
> 
> What's the purpose for having both an ENABLE_TESTS and a BUILD_TESTS option?

From what I recall `${PROJECT_NAME}_ENABLE_TESTS` is set by either the `find_package` or the user through cmake command line (ie `-Dswiftlets_ENABLE_TESTS=true`). The `${PROJECT_NAME}_BUILD_TESTS` is setup by the `SwiftCmakeOptions.cmake` based on the value of `${PROJECT_NAME}_ENABLE_TESTS`, which is normally a one to one assignment but we can add further logic to it (don't know if we do anything like that right now). We could for instance ask the system to build all tests across all organization and override the `${PROJECT_NAME}_ENABLE_TESTS` value sent in `find_package`, this might be useful if we wanted to compile all tests across our codebase and get a real sense of how well our code is covered across the board. The `${PROJECT_NAME}_BUILD_TESTS` is used by the build system to determine whether to include a test folder for example, we shouldn't be using `${PROJECT_NAME}_ENABLE_TESTS` in our CMakeLists.txt file.

FindAlbatross.cmake
isaactorz: Gotcha, thanks for clarifying.

PR #113, opened 2022-05-07
Removed warning messages (Option #2)

RReichert:
# Background

This PR is one of two options (see other option https://github.com/swift-nav/cmake/pull/112)

Please read https://github.com/swift-nav/cmake/pull/112 first as it is the preferred option and it outlines the problem.

### DIFF
--- a/FindAlbatross.cmake
+++ b/FindAlbatross.cmake
@@ -12,10 +12,8 @@
 
 include("GenericFindDependency")
 
-option(albatross_ENABLE_DOCS "" false)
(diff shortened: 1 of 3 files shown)


woodfell:
This option is much more in keeping with the spirit of this work originally since it was designed to cater to many different environments and flagging misuse of features was a big deal. That may or may not be the case any more but this one still feels like the "best" option because it only uses options which a package actually exports.

Another way to do this would be to update GenericFindDependency so that all these options default to off when a project is imported as a submodule but still default to on when used as a top level project

PR #114, opened 2022-05-12
Upgrade Sonarcloud Functionality

RReichert:
# Changes

I've now introduced a new function `generate_sonarcloud_project_properties` which no longer requires engineers to explicitly update the [sonar-project.properties](https://github.com/swift-nav/starling/blob/6039dfaca7c42984b6d38c1463f1f0efb7dcc2e9/sonar-project.properties) file whenever they introduce a new project or testing folder. This body of work now means as long as engineers call the `swift_add_executable`, `swift_add_test` and/or `swift_add_library` function, sonarcloud will now know which source/test files to analyze. It will exclude all other files which are deemed non production code.

# Future Work / Justification

Look we could have just kept the `sonar-project.properties` file system as it was, but in all honesty that lead to engineers forgetting to update the sonar cloud properties files (see [here](https://github.com/swift-nav/starling/pull/6558) for a list of missed projects). These changes also means that we can easily now develop a system which can correctly calculate the true code coverage for our Swift products by running all unit tests from Starling down to libswiftnav and aggregating the results into a code coverage report (not currently possible but the development of this system is easier now that we have this).

# Testing

* https://github.com/swift-nav/starling/pull/6558
* https://github.com/swift-nav/orion-engine/pull/3824


Sonarcloud.cmake
```diff
@@ -0,0 +1,215 @@
...
+#
+#   generate_sonarcloud_project_properties(${CMAKE_BINARY_DIR}/sonar-project.properties)
+#
+# This will generate a sonar project file in your root build directory (please
+# don't ever write this to your source directory). Prior to uploading the
+# results to Sonarcloud, you will need to manually transform all absolute path
+# references to relative paths. This manual step can be done easily with the
+# following Linux command:
+#
+#   sed -i -E "s|^(\s+)(${PWD}/)|\1|g" 'sonar-project.properties'
```
isaactorz, line 44: I know this is a one liner but I think it would be good to make it available as a bash script, and have pipelines use the bash script to fix up the paths rather than writing this snippet directly into pipeline scripts.

isaactorz (commented):
Looks

Sonarcloud.cmake
RReichert: This functionality is already incorporated into our Sonarcloud upload script ([see here](https://github.com/swift-nav/ci-jenkins/pull/590)). You can [see here](https://github.com/swift-nav/starling/pull/6558/files#diff-570fc702ebd2c9c0d8deef3f4063d278e1149f90679f8ef672ee565e262940eeR38-R41) how we use it.

isaactorz (approved):
lgtm this is a really nice capability you've introduced!

PR #117, opened 2022-06-24
Add Module For Checking Atomic

isaactorz:
Adds a cmake function that checks whether linking against libatomic is necessary for the toolchain in use.


isaactorz:
@RReichert What are the advantages of a C based version?

I can still look into implementing that also.

RReichert (approved):

RReichert:
> @RReichert What are the advantages of a C based version?
> 
> I can still look into implementing that also.

Just in case we want to use atomic in C programs, which we currently don't have a need for, so its not necessary.

PR #122, opened 2022-07-22
Add runtime sanitizer configuration file

peddie:
This was available in starling long ago but hasn't made it to other
repos.

### DIFF
--- a/SanitizeTargets.cmake
+++ b/SanitizeTargets.cmake
@@ -1,0 +1,94 @@
+# Runtime analysis using Clang sanitization flags.
+
+option(SWIFT_SANITIZE_ADDRESS "Enable address sanitizer." OFF)
+option(SWIFT_SANITIZE_LEAK "Enable leak sanitizer." OFF)
+option(SWIFT_SANITIZE_MEMORY "Enable memory sanitizer." OFF)
+option(SWIFT_SANITIZE_THREAD "Enable thread sanitizer." OFF)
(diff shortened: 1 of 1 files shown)


SanitizeTargets.cmake
```diff
@@ -0,0 +1,94 @@
+# Runtime analysis using Clang sanitization flags.
+
+option(ENABLE_SANITIZERS "Enable sanitizers." OFF)
```
isaactorz, line 3: Small nit but do we need this option? I think the way it's set up currently if I wanted to turn on UB sanitizing for instance i'd always have to do:
```sh
cmake -DENABLE_SANITIZERS=ON -DSANITIZE_UNDEFINED=ON ..
```

Seems a little redundant.

isaactorz (commented):
Another nit

Maybe we should prefix these options with `SWIFT` or something?

Would reduce the risk of colliding with similarly named options in other codebases.

SanitizeTargets.cmake
peddie: Sure, I'll remove the top-level flag.  I just took the file from starling.

peddie:
Seems a bit of a shame to stick `SWIFT` in front of everything, but on the other hand, I could see myself wasting a lot of time on subtle conflicts with a submodule somewhere without it.  Done

isaactorz (approved):

PR #123, opened 2022-07-22
Module to automatically fix include guards [ESD-2468]

woodfell:
A small script which can perform some linting on header files:
- Check that include guard matches the required pattern based on location in the source tree
- Add a copyright banner if missing

Changes are applied in place to any files

Added a cmake module which creates 2 new targets
- fix-include-guards - Run the python script to fix any errors
- fix-include-guards-check - Similar to clang-format-check, runs the script and then checks if any changes were applied returning an exit code if so. Can be used to fail a CI lint/format checking stage

python script is a cut down version of cpplint, removing all the stuff which isn't really relevant and adding the section for copyright notices

### DIFF
--- a/IncludeGuards.cmake
+++ b/IncludeGuards.cmake
@@ -1,3 +1,33 @@
+#
+# Copyright (C) 2021 Swift Navigation Inc.
+# Contact: Swift Navigation <dev@swift-nav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+#
+
+#
+# A module to check header files for:
+# - Include guards
+# - Copyright notice
+#
+# When included will create 2 new targets:
+# - fix-include-guards - Check all header files in the repository for the
+#   above conditions
+# - fix-include-guards-check - Depends on fix-include-guards then exits with
+#   an error code if any changes were applied. Useful for enforcing checks
+#   during CI runs
+#
+# Include guards are formatted according to the header file's location in the
+# source tree. Copyright notices are boiler plate block comments
+#
+# Changes are applied in place
+#
+
 add_custom_target(fix-include-guards-${PROJECT_NAME}
   COMMAND
     ${CMAKE_CURRENT_LIST_DIR}/scripts/fix_include_guards.py `git ls-files '*.h'`
--- a/IncludeGuards.cmake
+++ b/IncludeGuards.cmake
@@ -1,0 +1,14 @@
+add_custom_target(fix-include-guards-${PROJECT_NAME}
+  COMMAND ${CMAKE_CURRENT_LIST_DIR}/scripts/fix_include_guards.py `git ls-files '*.h'`
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  )
+add_custom_target(fix-include-guards-check-${PROJECT_NAME}
+  COMMAND git diff --exit-code
+  DEPENDS fix-include-guards-${PROJECT_NAME}
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  )
+
+if(${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
+  add_custom_target(fix-include-guards DEPENDS fix-include-guards-${PROJECT_NAME})
+  add_custom_target(fix-include-guards-check DEPENDS fix-include-guards-check-${PROJECT_NAME})
+endif()
--- a/IncludeGuards.cmake
+++ b/IncludeGuards.cmake
@@ -1,14 +1,21 @@
 add_custom_target(fix-include-guards-${PROJECT_NAME}
-  COMMAND ${CMAKE_CURRENT_LIST_DIR}/scripts/fix_include_guards.py `git ls-files '*.h'`
+  COMMAND
+    ${CMAKE_CURRENT_LIST_DIR}/scripts/fix_include_guards.py `git ls-files '*.h'`
   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-  )
+)
 add_custom_target(fix-include-guards-check-${PROJECT_NAME}
   COMMAND git diff --exit-code
   DEPENDS fix-include-guards-${PROJECT_NAME}
   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-  )
+)
 
 if(${PROJECT_NAME} STREQUAL ${CMAKE_PROJECT_NAME})
-  add_custom_target(fix-include-guards DEPENDS fix-include-guards-${PROJECT_NAME})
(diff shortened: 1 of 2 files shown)


IncludeGuards.cmake
```diff
@@ -0,0 +1,51 @@
+#
+# Copyright (C) 2021 Swift Navigation Inc.
```
isaactorz, line 2: nit: 2022

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */"""
+
+    # We'll say it should occur by line 10. Don't forget there's a
+    # placeholder line at the front.
+    for line in range(1, min(len(lines), 11)):
+        if re.search(r"Copyright", lines[line], re.I):
+            break
+    else:  # means no copyright line was found
```
isaactorz, line 122: Is this `else` properly aligned with the if statement?

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
```
RReichert, line 1: :clap: one step toward moving away from python `2.7.13` in CI

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+#
+
+import codecs
+import sys
+import re
+
+
+def FixHeaderGuard(filename):
```
RReichert, line 20: nit: python convention for function name is to use snake_case (pun intended). see https://peps.python.org/pep-0008/#function-and-variable-names.

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+    try:
+        with codecs.open(filename, "r", "utf8", "replace") as target_file:
+            lines = target_file.read().split("\n")
```
RReichert, line 23: Believe you can get away with replacing this line with `lines = [line.rstrip() for line in target_file.readlines()]`. that will mean that the next few lines 27-29 can be removed.

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Don't check for header guards if there are error suppression
+    # comments somewhere in this file.
+    #
+    # Because this is silencing a warning for a nonexistent line, we
+    # only support the very specific NOLINT(build/header_guard) syntax,
+    # and not the general NOLINT or NOLINT(*) syntax.
+    raw_lines = lines
+    for i in raw_lines:
+        if re.match(r"//\s*NOLINT\(build/header_guard\)", i):
```
RReichert, line 43: do we currently have something like this in our code base?

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+            return
+
+    # Allow pragma once instead of header guards
+    for i in raw_lines:
+        if re.match(r"^\s*#pragma\s+once", i):
```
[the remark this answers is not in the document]
RReichert: we need to support pedantic compilers like Greenhill which might not support it.

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+    #
+    # Because this is silencing a warning for a nonexistent line, we
+    # only support the very specific NOLINT(build/header_guard) syntax,
+    # and not the general NOLINT or NOLINT(*) syntax.
+    raw_lines = lines
+    for i in raw_lines:
+        if re.match(r"//\s*NOLINT\(build/header_guard\)", i):
+            return
+
+    # Allow pragma once instead of header guards
```
RReichert, line 46: would be good to revisit this later on since from memory out coding standards forbids this

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+
+    if re.match(r".*/include/.*", filename):
+        expected_guard = re.sub(r"^.*/include/", "", filename)
+    elif re.match(r"^include/.*", filename):
+        expected_guard = re.sub(r"^include/", "", filename)
+    else:
+        # header is probably just loose in a source directory so we'll have to use
+        # something else
+
+        components = filename.split("/")
```
RReichert, line 59: nit: libsbp is probably going to include these fix targets? should we support having them run on windows machine? if so this split might be a problem for them `os.path(...)` might be of use here.

scripts/fix_include_guards.py
```diff
@@ -0,0 +1,136 @@
...
+ * Copyright (C) 2022 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */"""
```
RReichert, line 115: change to:
```
    default_copyright = """/**
 * Copyright (C) {} Swift Navigation Inc.
 * Contact: Swift Navigation <dev@swiftnav.com>
 *
 * This source is subject to the license found in the file 'LICENSE' which must
 * be be distributed together with this source. All other rights reserved.
 *
 * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
 * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 */""".format(datetime.date.today().year)
```
Avoids having to update this every year :grin:

RReichert (approved):
nice, glad this is going in, will certainly help our code base keep that level of compliance. speaking of compliance, worth potentially looking at using [cpplint](https://github.com/cpplint/cpplint) for other things that we need to comply with.

scripts/fix_include_guards.py
isaactorz: [mindblown.gif](https://giphy.com/gifs/producthunt-mind-blown-blow-your-26ufdipQqU2lhNA4g)

lucasle-sn: I think it will make more sense passing the year of creation instead of current time.

lucasle-sn (approved):
LGTM

lucasle-sn (commented):
Sometimes I found incorrect comment after `endif`, which is also worth to be considered.

scripts/fix_include_guards.py
woodfell: Not sure, to be honest this entire script is just a cut down version of cpplint. I can't really take responsibility for most of it, I've just modified it to work with what we currently have

scripts/fix_include_guards.py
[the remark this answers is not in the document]
woodfell: Turned this in to an error

scripts/fix_include_guards.py
woodfell: Yeah perhaps, I'm going to leave this for the moment since (at least initially) it will only be run on developer laptops (linux, mac) or in a CI stage

PR #125, opened 2022-07-29
Remove checks introduced between clang tidy 6 and 14

krisukox:
This PR:
- removes few checks introduced between clang tidy 6 and 14 in order to enable bumping clang-format/clang-tidy to 14.
- adds the `WITHOUT_SWIFT_TYPES` option to the `swift_create_clang_tidy_targets` function

Tested here:
- https://github.com/swift-nav/starling/pull/6869
- https://github.com/swift-nav/orion/pull/4141

### DIFF
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -69,6 +69,8 @@
 # will explicitly disable these targets from the command line at configure time
 #
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE BOOL "Export compile commands" FORCE)
+
 # Helper function to actually create the targets, not to be used outside this file
 function(create_clang_tidy_targets key fixes)
   add_custom_target(
@@ -114,11 +116,6 @@ function(swift_create_clang_tidy_targets)
   if(NOT ${PROJECT_NAME}_ENABLE_CLANG_TIDY)
     early_exit(STATUS "${PROJECT_NAME} clang-tidy support is DISABLED")
   endif()
-
-  # This is required so that clang-tidy can work out what compiler options to use for each file
-  set(CMAKE_EXPORT_COMPILE_COMMANDS
-      ON
-      CACHE BOOL "Export compile commands" FORCE)
 
   find_program(CLANG_TIDY NAMES clang-tidy-14 clang-tidy)
 
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -204,7 +204,7 @@ function(swift_create_clang_tidy_targets)
         -readability-avoid-const-params-in-decls
         -readability-non-const-parameter
         -readability-redundant-declaration
-        -readability-uppercase-literal-suffix
+        -readability-redundant-member-init
         # Disable all new checks introduced between clang-6 and clang-14
         -clang-analyzer-core.StackAddressEscape
         -clang-analyzer-cplusplus.NewDeleteLeaks
@@ -240,7 +240,7 @@ function(swift_create_clang_tidy_targets)
         -readability-magic-numbers
         -readability-qualified-auto
         -readability-redundant-access-specifiers
-        -readability-redundant-member-init
+        -readability-uppercase-literal-suffix
         -readability-use-anyofallof)
 
     # Final list of checks to enable/disable
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -137,7 +137,6 @@ function(swift_create_clang_tidy_targets)
         cppcoreguidelines*
         google*
         misc*
-        modernize*
         performance*
         readability*)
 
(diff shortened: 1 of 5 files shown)


ClangTidy.cmake
```diff
@@ -217,7 +253,7 @@ function(swift_create_clang_tidy_targets)
 # Automatically generated, do not edit
 # Enabled checks are generated from SwiftClangTidy.cmake
 Checks: \"${comma_checks}\"
-HeaderFilterRegex: '.*'
+HeaderFilterRegex: '!.*/third_party/.*'
```
isaactorz, line 256: Which of the third party libraries were causing issues?

I'm guessing that clang-tidy uses `compile_commands.json` to find the header files to analyze.

We get around this for compiler warnings by marking third party headers as system includes:
https://github.com/swift-nav/orion/blob/935962e5b327d488e47776563f041d223acf1847/cmake/FindGrpc.cmake#L8

I assumed clang-tidy would behave similarly because we don't have a need for this filter in other projects like starling but all of the orion dependencies I've checked so far mark their headers as system.

ClangTidy.cmake
```diff
@@ -229,6 +265,11 @@ AnalyzeTemporaryDtors: true
   # Only lint targets created in this repository. Later on we will create 2 targets: clang-tidy-all will lint all "core" targets, executables and libraries clang-tidy-world will
   # lint everything including test suites
   swift_list_compilable_targets(all_targets ONLY_THIS_REPO SWIFT_TYPES "executable" "library")
+  if(x_WITHOUT_SWIFT_TYPES)
+    swift_list_compilable_targets(all_targets_without_swift_types ONLY_THIS_REPO)
+    list(APPEND all_targets ${all_targets_without_swift_types})
+    list(REMOVE_DUPLICATES all_targets)
+  endif()
```
isaactorz, line 272: Might try this as it's a little more concise. 

```suggestion
  if (x_WITHOUT_SWIFT_TYPES)
    swift_list_compilable_targets(all_targets ONLY_THIS_REPO TYPES "EXECUTABLE" "STATIC_LIBRARY" "SHARED_LIBRARY")
  else()
    swift_list_compilable_targets(all_targets ONLY_THIS_REPO SWIFT_TYPES "executable" "library")
  endif()
```

ClangTidy.cmake
```diff
@@ -256,6 +297,7 @@ AnalyzeTemporaryDtors: true
     message(WARNING "No sources to lint for clang-tidy-all, that doesn't sound right")
   else()
     list(REMOVE_DUPLICATES all_abs_srcs)
+    list(FILTER all_abs_srcs EXCLUDE REGEX "pb.cc")
```
isaactorz, line 300: What were these files?

isaactorz (commented):
I'd also like to see these changes tested in Starling just as a precaution.

Edit: just saw the starling PR link!

ClangTidy.cmake
woodfell: Yeah clang tidy should be picking up -isystem from the compile commands. I'd suggest that this is the wrong way to do it, perhaps just a local cmake config issue? I think that `THIRD_PARTY_INCLUDES_AS_SYSTEM` is turned off by default.

ClangTidy.cmake
woodfell: Generated nanopb files? Used by real-time replay and fast start

ClangTidy.cmake
krisukox: You're right, it works with `THIRD_PARTY_INCLUDES_AS_SYSTEM`: https://github.com/swift-nav/orion/pull/4141

FindStarling.cmake
```diff
@@ -32,4 +32,5 @@ GenericFindDependency(
     starling-util
   SOURCE_DIR starling
   SYSTEM_HEADER_FILE "pvt_driver/runner/pvt_runner.h"
+  SYSTEM_INCLUDES
```
isaactorz, line 35: Interesting. Starling is already using these linting settings so I'm surprised its headers are causing linting issues.

ClangTidy.cmake
isaactorz: @krisukox looking at this again are we picking up test files now? Maybe it was better how we had it before. Probably want to exclude test files from linting.

FindStarling.cmake
krisukox: I had a problem with starling-replay headers in the orion-engine PR.
We don't want to update the linter in the orion-engine right now, so I reverted this change.

isaactorz (commented):
```cmake
find_program(CLANG_TIDY NAMES clang-tidy-6.0 clang-tidy)
```

We're going to want to change this from clang-tidy-6.0 to clang-tidy-14

isaactorz:
Did some light testing on mac and seems to work well.

I was also curious if clang-tidy-6.0 would be able to handle the new `.clang-tidy` file and that seems to work fine as well.

isaactorz (approved):
LGTM

isaactorz:
@RReichert @woodfell anything you'd like to add before we merge?

PR #127, opened 2022-08-04
Fix Sonarcloud sonar-project.properties

krisukox:
The purpose of this PR is to generate code coverage from the headers which are included only in the test cases.
This PR:
- adds test source files to the `sonar.inclusions` property
- excludes test source files from the final code coverage by adding `sonar.coverage.exclusions` property


Sonarcloud.cmake
```diff
@@ -199,13 +199,23 @@ function(generate_sonarcloud_project_properties sonarcloud_project_properties_pa
...
+    endif()
+  endforeach()
   list(JOIN source_files ",${_sonarcloud_newline}" sonar_sources)
   string(APPEND sonarcloud_project_properties_content "sonar.inclusions=${_sonarcloud_newline}${sonar_sources}\n")
 
   set(test_files ${test_source_files})
-  list(JOIN test_files ",${_sonarcloud_newline}" sonar_tests)
-  string(APPEND sonarcloud_project_properties_content "sonar.tests.inclusions=${_sonarcloud_newline}${sonar_tests}\n")
+  foreach(source_file ${source_source_files})
+    list(REMOVE_ITEM test_files ${source_file})
```
isaactorz, line 215: What's the purpose of this? There's intersection between these sets of files?

isaactorz (commented):
It looks good but I would like to see some tests that activate the coverage.

Sonarcloud.cmake
krisukox: Yes, e.g.: `swiftlets-core` library and `unit-test-swiftlets-harness` test target have `threaded_swiftlet.cc` source file.

isaactorz (approved):
LGTM let's give it a go.

PR #130, opened 2022-08-30
compile compile command cleanup

RReichert:
## Changes

The main bit that I've change was introduce a new python script that can be used by Helix QAC to strip off any source file which it doesn't need to analyze. I've updated the script from where it was originally implemented [here](https://github.com/swift-nav/orion-engine/pull/4314#discussion_r957552705) so that its extensible for further develop (say something like white listing).

I've also done a few develop debt changes which have been on my plate since I'm already updating cmake:

1. I've moved a number of custom find package scripts from [here](https://github.com/swift-nav/orion/tree/master/cmake) to a common place
2. I've corrected an error I've made earlier where the find_package of orion-engine throw up warning messages about missing example/test libs/documentations.

### DIFF
--- a/FindOrion-Engine.cmake
+++ b/FindOrion-Engine.cmake
@@ -12,10 +12,7 @@
 
 include("GenericFindDependency")
 
-option(orion-engine_ENABLE_DOCS "" false)
-option(orion-engine_ENABLE_EXAMPLES "" false)
 option(orion-engine_ENABLE_TESTS "" false)
-option(orion-engine_ENABLE_TEST_LIBS "" false)
 
 GenericFindDependency(
   TARGET orion_engine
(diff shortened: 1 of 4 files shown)


FindOrion-Engine.cmake
```diff
@@ -12,10 +12,7 @@
 
 include("GenericFindDependency")
 
-option(orion-engine_ENABLE_DOCS "" false)
-option(orion-engine_ENABLE_EXAMPLES "" false)
```
RReichert, line 16: these were generating warning messages

FindPrometheus-Cpp.cmake
```diff
@@ -0,0 +1,32 @@
+include("GenericFindDependency")
```
RReichert, line 1: moving these files from https://github.com/swift-nav/orion/tree/master/cmake to common location

isaactorz (approved):
lgtm

PR #132, opened 2022-09-09
Fix compile_commands.json generation in newer cmake versions [BUILD-286]

krisukox:


isaactorz:
> How does this fix things? Do we need to set this in the non-old clang tidy module?

Currently we sort rely on at least on of these modules being included so that the `compile_commands.json` file gets output. This just makes it so that any module that requires the file makes sure to explicitly set the option to emit it.

The titles probably not super accurate.

isaactorz (approved):
lgtm but I've tagged Rodrigo since he wanted this fix.

RReichert (approved):

PR #135, opened 2022-09-23
Add find module for Auk

IsakTjernberg:
Follow up from #133, now with `auk_ENABLE_TESTS` instead of `BUILD_TESTS`.


armallen (approved):
Nice!

PR #140, opened 2022-12-05
Fix nanopb version 0.4.6 [BUILD-380]

krisukox:
This option is needed to use the newest nanopb with protoc older than 3.6.0.

Tested here: https://github.com/swift-nav/starling/pull/7239

### DIFF
--- a/FindNanopb.cmake
+++ b/FindNanopb.cmake
@@ -12,6 +12,7 @@
 
 include("GenericFindDependency")
 option(nanopb_BUILD_GENERATOR "" OFF)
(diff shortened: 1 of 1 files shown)


silverjam:
@isaactorz @krisukox Is this still needed?

isaactorz:
> @isaactorz @krisukox Is this still needed?

I *think* but I might not be remembering correctly: this would be necessary in order to bump nanopb in cmake. But we ended up not bumping because it also required a cmake update (which would have broken compat with one of our yocto sdks).

I think we might leave this one open since we will soon be able to break compat since we're about to be doing the builds in bazel, and are unblocked from updating nanopb on the cmake side.

PR #141, opened 2022-12-08
Don't fail on empty diffs when running clang-format

peddie:


isaactorz (approved):

isaactorz:
Would be good to see a test PR or something.

peddie:
By a test PR, do you mean a PR adding a test to CI to check that this doesn't error, or a one-off PR showing that this has the desired behaviour in a downstream repo?

isaactorz:
> By a test PR, do you mean a PR adding a test to CI to check that this doesn't error, or a one-off PR showing that this has the desired behaviour in a downstream repo?

Sorry, just circling back to this. Definitely the latter.

peddie:
I'll try to think of the right situation.  This addresses a problem I observed locally (failure of clang-format-diff caused a script to stop), but I've not seen a failure in CI.

PR #142, opened 2022-12-14
Add a script supporting ratcheting for clang-tidy warnings

peddie:
This allows you to set up a simple whitelist file for numbers of each
clang-tidy check that are expected to fail and check that these
haven't increased.

### DIFF
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -35,6 +35,16 @@
 # target described above and then return an error code if any warning/errors
 # were generated
 #
+# If the variable SWIFT_CLANG_TIDY_RATCHET_FILE is defined, then a
+# third target will be created: 'clang-tidy-${target}-ratchet-check'.
+# This target pipes the output of clang-tidy through
+# scripts/clang_tidy_ratchet.py and compares the output to the check
+# counts supplied in the filename in SWIFT_CLANG_TIDY_RATCHET_FILE,
+# returning failure only if the ratchet check script fails.  This
+# feature can be used to (approximately) mask existing clang-tidy
+# warnings without fixing them, so that development going forward
+# can still benefit from clang-tidy checking
+#
 # In addition there are two other targets created which lint multiple targets
 # at the same time
 #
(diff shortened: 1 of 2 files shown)


peddie:
See it in use in https://github.com/swift-nav/albatross/pull/378

ClangTidy.cmake
```diff
@@ -35,6 +35,16 @@
 # target described above and then return an error code if any warning/errors
 # were generated
 #
+# If the variable SWIFT_CLANG_TIDY_RATCHET_FILE is defined, then a
```
RReichert, line 38: worth adding that `SWIFT_CLANG_TIDY_RATCHET_FILE` must point to some file. Sorry I've never used ratchet file before

RReichert (approved):
I haven't had a very strong look over things, but I don't see anything troubling here.

ClangTidy.cmake
peddie: The rest of the comment explains how the value of that variable is used.

This is not something I would expect you to be familiar with -- I just made it up and wrote it down for this PR!

peddie:
A clarification on the intent of this change:

Independent of which checks are enabled, several downstream repos from this one have `clang-tidy` entirely turned off in CI, because as it stands, we don't have a way to make CI fail on new errors and pass on existing ones, for checks that we do want to enable and eventually completely fix.  Without this patch, we cannot flag new static analyzer errors in CI without fixing 100% of the existing errors.

The script in this patch approximates "fail on new errors and pass on existing ones" using counts, which is the most robust approach I found that wouldn't take days to implement correctly.  With this script, we can once again begin gating CI on `clang-tidy` checks and only get CI failures when we introduce new problems; we can then separately burn down the backlog of existing static analyzer failures.

My goal is not to be able to mask out checks that don't make sense for a given team or repository.  Changes to the checks we enable and disable are a different set of questions that I can't answer on my own.

isaactorz (commented):
I think this is great. Python script was really well organized.

One thing though (maybe I'm incorrect). It looks like there is an implicit 1-1 mapping between the ratchet file and a specific build target. And since you can only specify a single ratchet file, you can only lint a single build target per repo.

Do you think that will be sufficient in bigger projects?

woodfell (approved):

peddie:
> Do you think that will be sufficient in bigger projects?

No, probably not.  The way I've used this downstream is to wire CI to run `clang-tidy-all-ratchet-check` on the whole repo.  This is also how I've used this mechanism in the past in a monorepo.  But it could be nice for turnaround time to be able to check just a single target.

That mapping occurs entirely in cmake.  If you'd like, I can try to figure out how to break it down by target.  Personally, I'd probably rather wait until I can do it in starlark.

isaactorz:
> 



> > Do you think that will be sufficient in bigger projects?
> 
> No, probably not. The way I've used this downstream is to wire CI to run `clang-tidy-all-ratchet-check` on the whole repo. This is also how I've used this mechanism in the past in a monorepo. But it could be nice for turnaround time to be able to check just a single target.
> 
> That mapping occurs entirely in cmake. If you'd like, I can try to figure out how to break it down by target. Personally, I'd probably rather wait until I can do it in starlark.

My misunderstanding. I didnt' realize this was also creating an `all` target.

lgtm!

PR #143, opened 2022-12-20
Only set sanitizer suppression file if sanitzing is enabled

peddie:


peddie:
Not having this caused builds to fail on `clang` too old to support sanitizer suppression.

isaactorz (approved):

PR #144, opened 2023-01-10
Add FindMKL.cmake library[UTF-106]

john-michaelburke:
Adds a multi OS FindMKL library from this source:
https://github.com/marian-nmt/marian-dev/blob/master/cmake/FindMKL.cmake

I made one modification per the eigen doc's suggestion which was using `-DMKL_LP64` to set force the index type.

I have validated this is working as intended here:
https://github.com/swift-nav/orion-engine/pull/5217

### DIFF
--- a/FindMKL.cmake
+++ b/FindMKL.cmake
@@ -1,0 +1,137 @@
+################################################################################
+#
+# \file      cmake/FindMKL.cmake
+# \author    J. Bakosi
+# \copyright 2012-2015, Jozsef Bakosi, 2016, Los Alamos National Security, LLC.
+# \brief     Find the Math Kernel Library from Intel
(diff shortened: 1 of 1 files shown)


isaactorz (commented):
So it looks like unfortunately they're not using modern cmake best practices to create a target to link against. They're using the legacy mechanisms which is no longer recommended.

This is an example of a Find script using up to date best practices: https://github.com/swift-nav/orion-engine/blob/master/cmake/FindnetCDFCxx.cmake

isaactorz (commented):
Haven't been able to get around to sorting the build issues yet.

I do have a request however. Assuming no other project is going to need this dependency, could you move it to the orion-engine repo? It will make life a lot easier from a submodule handling perspective, considering the `cmake` project is our most common dependency.

PR #147, opened 2023-03-02
Add official MKL cmake finder.

john-michaelburke:
* Adds a stripped-down version of the official `${MKLROOT}/lib/cmake/mkl/MKLConfig.cmake`. 
* Attempts to find three static libraries off of known path MKLROOT
```
-- Found MKL: /opt/intel/oneapi/mkl/latest  
-- Found MKL: /opt/intel/oneapi/mkl/latest/lib/intel64/libmkl_intel_lp64.a  
-- Found MKL: /opt/intel/oneapi/mkl/latest/lib/intel64/libmkl_gnu_thread.a  
-- Found MKL: /opt/intel/oneapi/mkl/latest/lib/intel64/libmkl_core.a
```
* If MKL_ROOT it will first check to see if MKLROOT is defined, then try to fall back on a known location before failing.
* Using MKLROOT we deduce the include folder.
* Here is an example of how this would be called in orion-engine:
```
if(ENABLE_MKL)
  find_package(MKL)
  if (MKL_FOUND)
    foreach(TARGET IN LISTS TESTS-OE EXE-OE LIBS-OE)
      target_link_libraries(${TARGET} PRIVATE MKL::MKL)
    endforeach()
  endif()
endif()
```

### DIFF
--- a/FindMKL.cmake
+++ b/FindMKL.cmake
@@ -1,0 +1,850 @@
+#===============================================================================
+# Copyright 2021-2022 Intel Corporation.
+#
+# This software and the related documents are Intel copyrighted  materials,  and
+# your use of  them is  governed by the  express license  under which  they were
+# provided to you (License).  Unless the License provides otherwise, you may not
+# use, modify, copy, publish, distribute,  disclose or transmit this software or
+# the related documents without Intel's prior written permission.
+#
+# This software and the related documents  are provided as  is,  with no express
+# or implied  warranties,  other  than those  that are  expressly stated  in the
+# License.
+#===============================================================================
+
+#===================================================================
+# CMake Config file for Intel(R) oneAPI Math Kernel Library (oneMKL)
+#===================================================================
+
+#===============================================================================
+# Input parameters
+#=================
+#-------------
+# Main options
+#-------------
+# MKL_ROOT: oneMKL root directory (May be required for non-standard install locations. Optional otherwise.)
+#    Default: use location from MKLROOT environment variable or <Full path to this file>/../../../ if MKLROOT is not defined
+# MKL_ARCH
+#    Values:  ia32 intel64
+#    Default: intel64
+# MKL_LINK
+#    Values:  static, dynamic, sdl
+#    Default: dynamic
+#       Exceptions:- DPC++ doesn't support sdl
+# MKL_THREADING
+#    Values:  sequential,
+#             intel_thread (Intel OpenMP),
+#             gnu_thread (GNU OpenMP),
+#             pgi_thread (PGI OpenMP),
+#             tbb_thread
+#    Default: intel_thread
+#       Exceptions:- DPC++ defaults to tbb, PGI compiler on Windows defaults to pgi_thread
+# MKL_INTERFACE (for MKL_ARCH=intel64 only)
+#    Values:  lp64, ilp64
+#       GNU or INTEL interface will be selected based on Compiler.
+#    Default: ilp64
+# MKL_MPI
+#    Values:  intelmpi, mpich, openmpi, msmpi, mshpc
+#    Default: intelmpi
+#-----------------------------------
+# Special options (OFF by default)
+#-----------------------------------
+# ENABLE_BLAS95:      Enables BLAS Fortran95 API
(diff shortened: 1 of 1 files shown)


isaactorz (commented):
So the good thing about this version is they're using modern cmake conventions.

I do want to ask however, is the version we worked on not working?

This script is almost 900 lines of really complicated configurations. The goal of it is to "just work" for a huge number of potential target platforms.

However, for our needs, we probably only need to target one or two platforms and compilers. So I'd prefer if we checked in a simpler script that only targets our use cases.

We'll need to eventually port this to bazel and it'll go much smoother if we know exactly what options and settings we're building this library with, instead of having to filter through all this noise.

john-michaelburke:
> So the good thing about this version is they're using modern cmake conventions.
> 
> I do want to ask however, is the version we worked on not working?
> 
> This script is almost 900 lines of really complicated configurations. The goal of it is to "just work" for a huge number of potential target platforms.
> 
> However, for our needs, we probably only need to target one or two platforms and compilers. So I'd prefer if we checked in a simpler script that only targets our use cases.
> 
> We'll need to eventually port this to bazel and it'll go much smoother if we know exactly what options and settings we're building this library with, instead of having to filter through all this noise.

What we had was working in CI for orion-engine but not for orion (although it seemed like they were using the same Dockerfile which is puzzling). I was hoping this could also be more friendly for local development than what we spun up. As for porting to bazel, I see that Tensorflow supports mkl in their bazel builds so hopefully, we can reuse if not just copy their implementation for our usecase. In the meantime, I can take a stab at gutting some of this config.

FindMKL.cmake
```diff
@@ -0,0 +1,890 @@
...
+    target_compile_options(MKL::MKL_DPCPP INTERFACE ${MKL_DPCPP_COPT})
+    target_link_libraries(MKL::MKL_DPCPP INTERFACE ${MKL_LINK_LINE} ${MKL_THREAD_LIB} ${MKL_SUPP_LINK})
+    list(APPEND LINK_TYPES MKL::MKL_DPCPP)
+  endif()
+  # Single target for all C, Fortran link-lines
+  add_library(MKL::MKL INTERFACE IMPORTED GLOBAL)
+  target_compile_options(
+    MKL::MKL INTERFACE $<$<STREQUAL:$<TARGET_PROPERTY:LINKER_LANGUAGE>,C>:${MKL_C_COPT}> $<$<STREQUAL:$<TARGET_PROPERTY:LINKER_LANGUAGE>,Fortran>:${MKL_F_COPT}>
+                       $<$<STREQUAL:$<TARGET_PROPERTY:LINKER_LANGUAGE>,CXX>:${MKL_CXX_COPT}> $<IF:$<BOOL:${ENABLE_OMP_OFFLOAD}>,${MKL_OFFLOAD_COPT},>)
+  target_link_libraries(MKL::MKL INTERFACE ${MKL_LINK_LINE} ${MKL_THREAD_LIB} ${MKL_SUPP_LINK})
```
isaactorz, line 845: My suggestions would be to capture whatever these lines end up expanding to for our target and hardcoding that.

john-michaelburke: Sorry just saw this now. I'll add that after the hackathon.

john-michaelburke: @isaactorz I have addressed your concerns and hardcoded these blocks.

isaactorz (approved):
Just one small nit otherwise this is a lot clearer as to what options we're actually setting up in our build 👍🏼 .

FindMKL.cmake
```diff
@@ -0,0 +1,166 @@
...
+if(${CMAKE_VERSION} VERSION_LESS "3.13")
+  message(FATAL_ERROR "The minimum supported CMake version is 3.13. You are running version ${CMAKE_VERSION}")
+endif()
+
+include(FindPackageHandleStandardArgs)
+
+set(MKLROOT_DEFAULT "/opt/intel/oneapi/mkl/latest")
+set(MKL_ARCH intel64)
+set(MKL_INTERFACE_FULL intel_lp64)
+set(MKL_THREADING gnu_thread)
```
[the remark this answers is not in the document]
john-michaelburke: Done. Fair enough.

PR #150, opened 2023-03-27
Turn off duplicate detection in SonarCloud for test targets [SEN-798]

reimerix:
Turning off code duplication warnings for test targets. Tested in Starling here: https://github.com/swift-nav/starling/pull/7760

This has been getting into our way a lot, since we want to make most tests [DAMP rather than DRY](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html).


isaactorz (approved):
lgtm!

silverjam:
@isaactorz Do we need to log a Jira ticket to implement this feature in Bazel?

isaactorz:
> @isaactorz Do we need to log a Jira ticket to implement this feature in Bazel?

I'll add it to the ticket that tracks bazel integration with sonarcloud.

PR #152, opened 2023-05-12
add FindBoringSSL.cmake

wwieclaw:
# Testing
Tested in related PRs.

# JIra
https://swift-nav.atlassian.net/browse/BUILD-679

# Related PRs
* https://github.com/swift-nav/orion_proto/pull/1501
* https://github.com/swift-nav/orion/pull/5748

### DIFF
--- a/FindBoringSSL.cmake
+++ b/FindBoringSSL.cmake
@@ -16,4 +16,3 @@ GenericFindDependency(
   SOURCE_DIR boringssl
   SYSTEM_INCLUDES
 )
-
--- a/FindBoringSSL.cmake
+++ b/FindBoringSSL.cmake
@@ -1,0 +1,19 @@
+#
+# Copyright (C) 2021 Swift Navigation Inc.
+# Contact: Swift Navigation <dev@swift-nav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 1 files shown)


FindBoringSSL.cmake
```diff
@@ -0,0 +1,18 @@
...
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+#
+
+include("GenericFindDependency")
+GenericFindDependency(
+  TARGET ssl
```
isaactorz, line 15: nit: I does boringssl create an alias target? i.e. `boringssl::ssl`

If so I would prefer we use those versions as `ssl` is really generic and cmake doesn't have any sort of namespacing.

FindBoringSSL.cmake
```diff
@@ -0,0 +1,18 @@
...
+  SOURCE_DIR boringssl
```
isaactorz, line 16:
```suggestion
  ADDITIONAL_TARGETS crypto
  SOURCE_DIR boringssl
```

Lets add `ADDITIONAL_TARGETS` so we also make sure that the crypto library was found and configured correctly.

FindBoringSSL.cmake
wwieclaw: Yeah I was dissapointed to find out that it doesn't

isaactorz (approved):
lgtm!

PR #153, opened 2023-05-15
[ASILSAFE-150] Upgrade warnings to errors option in clang tidy generator

RReichert:
## Changes

For ASIL projects, we want to upgrade all clang-tidy warnings to errors. The clang-tidy setup in `orion-engine` is useless and doesn't flag anything. This work allows us to upgrade the checks in orion when needed.

### DIFF
--- a/ClangTidy.cmake
+++ b/ClangTidy.cmake
@@ -112,7 +112,7 @@ function(swift_create_clang_tidy_targets)
     return()
   endif()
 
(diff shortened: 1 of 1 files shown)


woodfell (approved):

woodfell:
Change looks fine, but isn't this what the `clang-tidy-check` target is for? It should return an error (and therefore fail the build when run by Jenkins) if any warning are generated

RReichert:
:facepalm: yeah I can drop this PR than, I didn't realize that that does the same thing. thanks @woodfell

PR #159, opened 2023-07-13
Don't fail sanitizer builds on failed optimizations

peddie:


RReichert (commented):
mind explaining why this is necessary and how you got to this issue? I don't see this as a positive move forward.

peddie:
Sorry, I have a comment in the Bazel counterpart, but I did not copy it here.

When building with sanitizers, the overhead of all the instrumentation inserted can cause some functions to get much larger.  Some grow so large that the compiler chooses to fall back to a simple register allocation scheme rather than the normal one based on graph optimization.  When this happens and `-Werror=disabled-optimization` is enabled, the build fails.  

In this situation, it's better to continue building and accept that the register allocation will be worse, in the interest of getting sanitizer results.  We could try hardcoding different values for options like `--param ira-max-conflict-table-size` for each compiler and hope we set them large enough, but that's fragile, and it will take even longer to build these functions.  Instead, we just instruct the compiler not to treat this situation as a fatal error.

This affects only a handful of functions, and only when they are built with sanitizers enabled.  There is no effect on production code.  Is there a downside you have in mind?

RReichert (approved):
Thanks for the explanation, made a lot of sense.

> There is no effect on production code. Is there a downside you have in mind?

When I see disabling warnings, it immediately raises alarm bells. So I wanted to make sure its a sensible reason, which it seems to be.

isaactorz (commented):
Thanks for that detailed explanation Matt. Lgtm.

isaactorz:
For some reason the UI is not giving me the `Approve` option. 😬

PR #160, opened 2023-07-19
Add exclude flag for doxygen

lucasle-sn:


RReichert:
@lucasle-sn do you have an associated PR which uses the cmake changes?

lucasle-sn:
> @lucasle-sn do you have an associated PR which uses the cmake changes?

This is the PR I am working on that temporarily needs to workaround doxygen check for a single file `fsm.h` https://github.com/swift-nav/starling-core/pull/314. However, I found there is also another place excluding files in `starling-core`. That's why I think it is worth adding it.

woodfell:
Parked based on discussions last week

PR #161, opened 2023-07-26
add CODEOWNERS

isaactorz:
Enables developer infrastructure to be aware of any development going on in the cmake repo.

Particularly we want to avoid new features being developed for cmake that are not accounted for in the bazel system.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+@swift-nav/devinfra


isaactorz:
cc @silverjam do we need to edit the project settings so that it requires dev-infra for reviews in this repo?

silverjam (approved):
Yes please

isaactorz:
> Yes please

I don't have permissions to view/edit the settings in this repo.

I thought just adding the `CODEOWNERS` file would make the reviews required but it doesn't look like that's the case.

PR #168, opened 2023-09-20
feat: clang-format: add GIT_DESCRIBE_MATCH

silverjam:


isaactorz (commented):
Does this make sense in this context?

The idea here is that the last tag is a heuristic for "get me all the files that have changed in this diff"

I don't think it hurts much to match on a pattern - there'll just be longer list of files that the formatter will be run against - but I also don't see how it's useful.

silverjam:
> Does this make sense in this context?
> 
> The idea here is that the last tag is a heuristic for "get me all the files that have changed in this diff"
> 
> I don't think it hurts much to match on a pattern - there'll just be longer list of files that the formatter will be run against - but I also don't see how it's useful.

Nope, doesn't make sense given that, thanks.

PR #169, opened 2023-10-04
Add iec559 check

armallen:
# Rationale

We need to check double type definition for ASIL purposes. This adds a simple check, then called by integrity_core_common (or any library needing to check double definition)

# Related PRs
https://github.com/swift-nav/rules_swiftnav/pull/115
https://github.com/swift-nav/orion-engine/pull/7226


isaactorz (commented):
I'd prefer if this applied at the target level instead of globally for the project.

When the project uses gcc/clang this hardly matters but I could see this being reused in contexts where that's not the case and the entirety of the codebase doesn't have a requirement for asil compliance.

PR #171, opened 2024-01-11
Handle new suitesparse version

AlttiJokinen:
https://swift-nav.atlassian.net/browse/OC-729

Handle directory structure of new Suitesparse version

### DIFF
--- a/Findsuitesparse.cmake
+++ b/Findsuitesparse.cmake
@@ -18,13 +18,13 @@ endif()
 find_path(suitesparse_cholmod_INCLUDE_DIR
   NAMES cholmod.h
   HINTS ${SUITESPARSE_DIRS}
-  PATH_SUFFIXES "suitesparse" "include/suitesparse"
+  PATH_SUFFIXES "include" "include/suitesparse"
 )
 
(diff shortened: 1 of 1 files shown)


Findsuitesparse.cmake
```diff
@@ -18,13 +18,13 @@ endif()
 find_path(suitesparse_cholmod_INCLUDE_DIR
   NAMES cholmod.h
   HINTS ${SUITESPARSE_DIRS}
-  PATH_SUFFIXES include
```
jbangelo, line 21: I think we need to keep the `include` suffix for the Linux installs and older versions on Mac

AlttiJokinen: It seems that I made mistake, it should have been include and not suitesparse

AlttiJokinen:
@ljbade  I can see that your approval is needed, because must be someone from dev infra

ljbade:
Has this now been tested on both Linux and Mac?

AlttiJokinen:
@ljbade  Yes, it has been compiled as part of Albatross pipeline https://github.com/swift-nav/albatross/pull/473#issuecomment-1888240490

ljbade (approved):

PR #172, opened 2024-02-07
Add FindOrion module

john-michaelburke:
- Adds FindOrion.cmake and hopes no one else needs to use it :skull: 
- We are working on some load testing scripts which may need access to some of the rpc code in orion for mocking purposes.

### DIFF
--- a/FindOrion.cmake
+++ b/FindOrion.cmake
@@ -1,0 +1,20 @@
+#
+# Copyright (C) 2024 Swift Navigation Inc.
+# Contact: Swift Navigation <dev@swift-nav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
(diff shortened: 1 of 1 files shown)


FindOrion.cmake
```diff
@@ -0,0 +1,21 @@
...
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+#
+
+include("GenericFindDependency")
+
+set(orion_ENABLE_CLANG_FORMAT ON)
```
john-michaelburke, line 15: Not sure if this is the most appropriate way to do this but this flag is required by orion so it seemed fitting to add it.

PR #173, opened 2024-03-25
enable c++20

isaactorz:
Adds a couple of functions that are essentially necessary to enable using libc++ and clang-15 with C++20

I tested locally on mac with clang-15 and libc++

### DIFF
--- a/CheckExperimentalMemoryResource.cmake
+++ b/CheckExperimentalMemoryResource.cmake
@@ -1,0 +1,28 @@
+include(CheckCXXSourceCompiles)
+
+# Some toolchains require explicitly linking against libatomic
+#
+# If such linking is required, then this function sets the
+# value of result to "atomic". Otherwise it remains untouched.
+# This makes it so that the function only needs to be called once
+# per project.
+#
+# It can be used as follows:
+#
+# check_cxx_needs_atomic(LINK_ATOMIC)
+# target_link_libraries(foo PRIVATE ${LINK_ATOMIC})
+# ...
+# target_link_libraries(bar PRIVATE ${LINK_ATOMIC})
+#
+function(check_experimental_memory_resource result)
+  check_cxx_source_compiles("
+#include <experimental/memory_resource>
+  int main() {
+    return 0;
+  }
+  " success)
+
+  if(success)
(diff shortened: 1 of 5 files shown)


pcrumley (approved):
Wait for @RReichert approval before merging plz

RReichert (approved):
Although the code works, I question if this is the best way forward. I would recommend something like this:

```cmake
function(check_libcpp success)
  check_cxx_source_compiles("
    #include <iostream>
    int a =
    #ifdef _LIBCPP_VERSION
      1;
    #else
      kdfasfdl
    #endif
    int main() {
      return 0;
    }
  " ${success})
endfunction()
```

Where the function returns the true/false results. Its the responsibility of the caller to than add the definition to the compiler via `target_compile_definitions`. Using `-D` command line option is compiler specific way to add definitions, other compilers might not follow that approach (even though everything we support currently does). 

I'd imagine that these checks would be used by one of the `swift_add_*` target functions to add the definitions to the source files, as `PUBLIC` definitions.

isaactorz:
> Although the code works, I question if this is the best way forward. I would recommend something like this:
> 
> ```cmake
> function(check_libcpp success)
>   check_cxx_source_compiles("
>     #include <iostream>
>     int a =
>     #ifdef _LIBCPP_VERSION
>       1;
>     #else
>       kdfasfdl
>     #endif
>     int main() {
>       return 0;
>     }
>   " ${success})
> endfunction()
> ```
> 
> Where the function returns the true/false results. Its the responsibility of the caller to than add the definition to the compiler via `target_compile_definitions`. Using `-D` command line option is compiler specific way to add definitions, other compilers might not follow that approach (even though everything we support currently does).

That was mostly for readability - `_LIBCPP_ENABLE_CXX20_REMOVED_FEATURES` would also work fine with `target_compile_definitions` - I just thought that it would make what we're doing more obvious for a less cmake savvy future readers. I figure it's unlikely we use this standard library impl with anything other than gcc/clang going forward.
> 
> I'd imagine that these checks would be used by one of the `swift_add_*` target functions to add the definitions to the source files, as `PUBLIC` definitions.

I agree with your overall point. It's not a good API - we're writing it to a specific implementation issue instead of providing an interface for the caller.

I've updated the functions to just be pass/fail and then the caller can figure out what to do with the result.

PR #178, opened 2024-09-05
CMake patches for QNX Support

adharsh:
Patches to support building Starling on QNX


ljbade (approved):

pcrumley:
will any of this work need to be ported/implemented for Bazel?

adharsh:
> will any of this work need to be ported/implemented for Bazel?
Yup! That's the plan, the goal is to first get the CMake build to work

PR #183, opened 2025-05-06
Update protobuf to v25.1

RReichert:
# Changes

Since we are upgrading the protobuf version, the location for the root CMakeLists.txt has changed, hence the update.

We also updated the swift targets (ex: `swift_add_library` or `swift_add_executable`) so that we can now set a global default C++ version with `SWIFT_CXX_STANDARD`.

Lastly tried to unify the interface between `swift_add_test` and the rest of the other swift targets since `swift_add_test` was developed separately to the rest, hence the difference in interface.

These changes will allow us to set the C++ for our Swift targets independently from the third party libraries.


FindProtobuf.cmake
```diff
@@ -10,13 +10,14 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 #
 
+option(ABSL_PROPAGATE_CXX_STD "Use CMake C++ standard meta features (e.g. cxx_std_14) that propagate to targets that link to Abseil" true)
```
RReichert, line 13: protobuf now has abseil, hence the need for this. I believe in the very latest version they've removed abseil, but we aren't there yet.

SwiftTargets.cmake
```diff
@@ -250,8 +250,11 @@ function(swift_add_target target type)
   if (x_C_STANDARD)
     list(APPEND language_standards_args C ${x_C_STANDARD})
   endif()
+
   if (x_CXX_STANDARD)
     list(APPEND language_standards_args CXX ${x_CXX_STANDARD})
+  elseif (SWIFT_CXX_STANDARD)
+    list(APPEND language_standards_args CXX ${SWIFT_CXX_STANDARD})
```
RReichert, line 257: here if the caller specified the `CXX_STANDARD` it prioritizes that over the global variable `SWIFT_CXX_STANDARD`

TestTargets.cmake
```diff
@@ -261,7 +261,7 @@ endfunction()
 
 function(swift_add_test target)
   set(argOption "INTEGRATION_TEST" "PARALLEL" "POST_BUILD" "UNIT_TEST" "VALGRIND_MEMCHECK")
-  set(argSingle "COMMENT" "WORKING_DIRECTORY")
+  set(argSingle "C_STANDARD" "CXX_STANDARD" "COMMENT" "WORKING_DIRECTORY")
```
RReichert, line 264: we can not set the C++ standard used by the test code via the function call

jbangelo (approved):

